PR #39, opened 2023-07-22
Fixed Transaction Mapping and Sellers bugs 

JustSamuel:
Fixed two bugs:  
- Bug where credit invoice entries contained products from SubTransactions with different sellers  
- Bug where transaction report reported products being sold from subTransactions from different sellers.


CodeNamedRobin (approved):
LGTM!

PR #49, opened 2023-08-30
Change NodeJS runner to add more workers & improve performance

Yoronex:
Should improve performance of SudoSOS and stop the POS from "hanging". It is the same fix as was implemented in the SNiC CelerIT backend. However, because SudoSOS also has timed tasks to synchronize stuff, these had to be moved to a separate file. Otherwise, each worker would start the tasks at exactly the same time and thus the same work would be done four times, defeating the point of having workers.

The docker container starts correctly on my machine, but I have not yet tested whether SudoSOS actually works correctly, as I do not have a production environment setup on my machine. Therefore, I propose to test this on sudosos.test.gewis.nl.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,12 +14,14 @@ RUN apk add openssl
 
 WORKDIR /app
 COPY ./package.json ./package-lock.json ./
-RUN npm install --production
+RUN npm ci
+RUN npm install pm2 -g
 ARG TYPEORM_USERNAME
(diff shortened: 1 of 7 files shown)


JustSamuel (changes requested):
Does this not cause the same issue as with celerit making to many connections to the database server?

JustSamuel (approved):
Hope it works lmao

CodeNamedRobin (approved):
LGTM!

PR #56, opened 2023-09-01
Fixed incorrect collation

CodeNamedRobin:
Somehow the collations were not correct, so we fixed them.


tomudding (approved):
This is the least intrusive way to fix this now.

It is preferred to determine whether `utf32` should really be used in the database instead of `utf8(mb4)`. As it does not really make sense imho.

PR #57, opened 2023-09-01
Remove collation from search queries

CodeNamedRobin:
The database has its charset set to `utf32` and the collation to `utf32_general_ci`. Forcing the search queries to collate with `utf8_general_ci` breaks, as this is not supported. The `COLLATE` statement is not necessary.

Because GH-56 did not work as expected. This code has already been tested and verified to work.


tomudding (approved):
It is preferred to determine whether `utf32` should really be used in the database instead of `utf8(mb4)`. As it does not really make sense imho.

Issue #67, opened 2023-09-18
Fine-related endpoints do not work well without a frontend


Yoronex:
The current fine operations are of insufficient quality. They work well in theory, but in practice the endpoints have some flaws which block the handout of fines. Therefore, these issues should be resolved with some haste.

From what I have heard from @CodeNamedRobin and also experienced myself when observing the usage of these endpoints, the current implementation has the following flaws which should be addressed:

1. The endpoints are not stateless. Some endpoints have an optional reference date, but when this is not used, a previous handout event date is picked. If this does not exist, the current date is picked. This makes using the endpoint very confusing without a frontend, as a user now has to think for themselves what the endpoint should and will do. Therefore, the reference date should be more explicit.
2. For handing out fines, there are actually two reference dates: the previous fine handout event and the date of the new handout event. In the current implementation, the latter reference date cannot be set and is always "now". This is an easier implementation, but makes it impossible to hand out fines the morning after the borrel when issues come up.
3. The endpoints only return IDs of users and not any names. This is fine when a frontend also has a complete list of all users, but this is impossible without it. Therefore, the responses should include the `BaseUserResponse` object to make it clear which user belongs to the given ID.
4. The `GET /fines/eligible` endpoint only gives a list of users and their fines, but does not include any balances. This is by design, because balances should be requested from the `GET /balances` endpoint. However, it is not possible to request a subset of users. The current endpoints only allow requesting one user or all users. Therefore, a filter should be added to the `GET /balances` endpoint to also provide a list of IDs the user wants the balances for.
5. The `GET /fines/eligible` and `POST /fines/handout` endpoints do not return the reference dates used. When the system "assumes" certain dates or parameters, these should be included in the responses to verify that the given result is expected and valid. However, given the flaws given above, there should be no more assumptions made by the system.
6. When a user goes out of debt, the `currentFines` attribute should be set to `null`. However, this does not seem to happen in the production environment. It might have something to do with the different database used (MySQL), but this should be investigated and fixed.

**Todo list to solve this issue**
- [x] Make referenceDate(s) explicit in calls to eligible, handout and notify endpoints
- [x] (Temporarily) include `BaseUserResponse` and balance on all reference dates in response of `GET /fines/eligible`.
- [x] Include reference date(s) in `FineHandoutEvent`.
- [x] Fix `User.currentFines` not being set to `null` when user pays off debt.

Yoronex:
[...]

PR #78, opened 2023-09-29
Fine design improvements

Yoronex:
Fixes https://github.com/GEWIS/sudosos-backend/issues/67.

@JustSamuel Can you also check whether the Swagger spec is consistent with the actual implementation? It should be the case, but a double check never hurts.

### DIFF
--- a/src/controller/debtor-controller.ts
+++ b/src/controller/debtor-controller.ts
@@ -186,14 +186,14 @@ export default class DebtorController extends BaseController {
 
     let params;
     try {
-      const referenceDates = asArrayOfDates(req.query.referenceDate);
+      const referenceDates = asArrayOfDates(req.query.referenceDates);
       if (referenceDates === undefined && req.query.referenceDates === undefined) throw new Error('referenceDates is required');
       if (referenceDates === undefined && req.query.referenceDates !== undefined) throw new Error('referenceDates is not a valid array');
       params = {
         userTypes: asArrayOfUserTypes(req.query.userTypes),
         referenceDates,
       };
-      if (params.userTypes === undefined) throw new Error('userTypes is not a valid array of UserTypes');
+      if (params.userTypes === undefined && req.query.userTypes !== undefined) throw new Error('userTypes is not a valid array of UserTypes');
     } catch (e) {
       res.status(400).json(e.message);
       return;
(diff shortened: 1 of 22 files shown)


src/controller/debtor-controller.ts
```diff
@@ -174,8 +174,9 @@ export default class DebtorController extends BaseController {
    * @group debtors - Operations of the debtor controller
    * @operationId calculateFines
    * @security JWT
-   * @param {Array<string>} userTypes[].query.required - List of all user types fines should be calculated for
-   * @param {string} referenceDate.query - Date to base fines on. If undefined, use the date of the last fine handout event. If that one also does not exist, use now.
+   * @param {Array<string>} userTypes[].query - List of all user types fines should be calculated for
```
JustSamuel, line 177: `Array.<string>`

src/controller/debtor-controller.ts
```diff
@@ -185,14 +186,16 @@ export default class DebtorController extends BaseController {
 
     let params;
     try {
+      const referenceDates = asArrayOfDates(req.query.referenceDates);
+      if (referenceDates === undefined && req.query.referenceDates === undefined) throw new Error('referenceDates is required');
+      if (referenceDates === undefined && req.query.referenceDates !== undefined) throw new Error('referenceDates is not a valid array');
```
JustSamuel, line 191: niet super duidelijk imho, want het komt over alsof deze twee regels altijd true zijn.

src/controller/response/debtor-response.ts
```diff
@@ -19,15 +19,18 @@ import { DineroObjectResponse } from './dinero-response';
...
 import { PaginationResult } from '../../helpers/pagination';
 import { BaseUserResponse } from './user-response';
+import BalanceResponse from './balance-response';
 
 /**
  * @typedef UserToFineResponse
  * @property {integer} id - User ID
- * @property {DineroObjectResponse.model} amount - Amount to fine
+ * @property {DineroObjectResponse.model} fineAmount - Amount to fine
+ * @property {Array.<BalanceResponse>} balances - Balances at the given reference dates
```
JustSamuel, line 28: should be .required but we have done this wrong almost everywhere lmao.

JustSamuel (approved):
small remarks

Issue #74, opened 2023-09-28
Borrelkaart users do not automatically accept TOS


CodeNamedRobin:
Title. This means that right now nothing can be ordered on behalf of their 'name'.

Issue #75, opened 2023-09-28
Borrelkaart users can have negative balance


CodeNamedRobin:
This is really annoying, as this would create debt that would never get paid off.

PR #79, opened 2023-09-29
Voucher auto ToS accept & restrict users who can go into debt

Yoronex:
Fixes https://github.com/GEWIS/sudosos-backend/issues/74 and https://github.com/GEWIS/sudosos-backend/issues/75.

Note that this pull request requires a (manual) migration, as by default users can no longer go into debt. For all currently active users, the `canGoIntoDebt` column has to be set to `true`.


src/controller/response/user-response.ts
```diff
@@ -41,6 +41,7 @@ export interface BaseUserResponse extends BaseResponse {
  * @property {boolean} extensiveDataProcessing - Whether data about this
  * user can be used (non-anonymously) for more data science!
  * @property {boolean} ofAge - Whether someone is old enough to drink beer
+ * @property {boolean} canGoIntoDebt - Whether this user can get a negative balance
```
JustSamuel, line 44: .required

src/controller/transaction-controller.ts
```diff
@@ -157,8 +157,7 @@ export default class TransactionController extends BaseController {
 
       // verify balance if user cannot have negative balance.
       const user = await User.findOne({ where: { id: body.from } });
-      const allowNegative = this.roleManager.can(await this.roleManager.getRoles(user), 'update', 'own', 'Balance', ['negative']);
-      if (!allowNegative && !await TransactionService.verifyBalance(body)) {
+      if (!user.canGoIntoDebt && !await TransactionService.verifyBalance(body)) {
```
JustSamuel, line 160: Is this preferred? Now going into debt is no longer role based, which seems less flexible.

JustSamuel (approved):
implementation is fine, still questioning if we do not want the RBAC. but looks fine,

src/controller/transaction-controller.ts
Yoronex: Currently, there are no mechanisms to dynamically assign roles to and remove roles from users. If we wanted to use RBAC, we have to create these. Also, when using RBAC, in theory multiple roles can define whether someone can go into debt. Finding out based on the list of roles whether one can go into debt is therefore much harder. Simply creating a boolean that is related to the user entity, is therefore in my opinion the most simple solution. We can then also reuse the existing `createUser` and `updateUser` endpoints.

Issue #70, opened 2023-09-18
Logging seems to be broken


CodeNamedRobin:
The logging in the backend seems to not work anymore. Even when the logger mode is set to ALL, not much seems to happen. E.g. transfers/transactions are not logged, hampering debugging.

Yoronex:
https://log4js-node.github.io/log4js-node/clustering.html Might be a possible fix.

PR #80, opened 2023-10-02
Fix trace logs missing from production container

Yoronex:
Should fix https://github.com/GEWIS/sudosos-backend/issues/70.


src/entity/user/user.ts
```diff
@@ -127,4 +127,15 @@ export default class User extends BaseEntity {
 
   @OneToMany(() => AssignedRole, (role) => role.user)
   public roles: AssignedRole[];
```
JustSamuel, line 129: clearly not part of this PR; but I'll allow it.

JustSamuel (approved):

PR #81, opened 2023-10-11
changed res from DELETE  to return empty object and 204 status code. 

JoHeyJo:
- Function docs also reflect changes.
- Test cases were not changed to reflect changes. Test for `DELETE /banners` will fail.


Yoronex:
Issue has been resolved in https://github.com/GEWIS/sudosos-backend/pull/129.

PR #86, opened 2023-10-14
Ensure that only LOCAL_USERS can get a reset password token.

JustSamuel:


src/controller/authentication-controller.ts
```diff
@@ -15,18 +15,18 @@
  *  You should have received a copy of the GNU Affero General Public License
  *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-import { Request, Response } from 'express';
-import log4js, { Logger } from 'log4js';
-import BaseController, { BaseControllerOptions } from './base-controller';
```
Yoronex, line 20: Why are all those spaces deleted here? If I remember correctly, the ESLint wants these spaces here? Weirdly enough the CI does not complain about it, but it is very inconsistent with the rest of the codebase.

package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sudosos-back-end",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
```
Yoronex, line 4: Please remove package-lock.json from this PR, as package.json itself has not changed.

Is this lock file bump caused by Node 20 btw? Because I only ever see lockfile version 2.

Yoronex (changes requested):

Issue #62, opened 2023-09-09
Add explicit POS authentication & authorization


Yoronex:
Right now, the used POS in SudoPOS is hardcoded (i.e. ID 1). This is undesirable, as now only owners of POS 1 can switch between POSs. This means that a POS from a different group of users with no overlap, cannot be switched to without adding someone from POS 1 to that specific group. Secondly, a POS can be opened anywhere in the world. Because PINs are quite easy to brute force, this is a major security risk (note that in current production, this vulnerability has been mitigated by having strict webproxy rules and only allowing traffic from within the GEWIS IP range to access SudoPOS and its corresponding endpoints).

To resolve these issues, we should implement a new method of authenticating a POS. The best way to do this (as discussed at the ABC meeting) would be to create a User entity for each POS, corresponding with that POS. This user can then perform actions in name of the POS, with allows for correct authorization and separation of concerns.

**Workflow**
1. When visiting the POS, the POS should check for the presence of a user token identifying the POS. If it does not exist, it should redirect to a "normal" login page.
2. A POS owner logs in with their credentials. They get a list of all POSs they are allowed to open.
3. When a user chooses a POS to open, the backend will authenticate the POS's corresponding User entity and send a token with a long lifespan (like 1 week) to the POS. The POS frontend should store this token in persistent storage.

**Pro's of this new method**
- API requests like authenticating with PIN, fetching users and fetching products can then be done by the POS User. General User tokens then no longer have to do this, meaning their access can be revoked (see also https://github.com/GEWIS/sudosos-backend/issues/23).
- To "give" a POS terminal to a different group, a POS owner can simply close their POS and thus allow someone else to log in and open their POS.
- PIN authentication is no longer open to the internet, but requires some form of authentication.

SuperVK:
TODO (copied from #232):

When the POS has implemented this new flow, we can start with improving the lesser tokens:
- Lesser tokens should include the ID of the POS they have been authenticated at. This ensures the user can only create transactions on the POS it is authenticated at. Furthermore, we then also know whether the user is allowed to switch the POS, because we know whether they own the current POS as well.
- PIN, NFC and EAN authentications should also require JWT tokens, namely the token of the POS.
- Revoke the RBAC rights for regular users to fetch all other users, POSs, containers, products, and product categories.

PR #94, opened 2023-10-17
Draft: Add a separate endpoint for requesting base users.

Kevinjil:
Regarding issue 23, this will allow more fine grained control over permissions to request specific user responses in the future.

As of now, there is still some implicit leaked information, as the search filters are used as-is, while the information on which it filters is excluded from responses.

TODO:

- [x] Implement endpoint for requesting base user responses
- [ ] Modify existing roles to reflect the separation in endpoints
- [ ] Final fix using #62


SuperVK:
The branches have grown too far apart and it seems unlikely that this will be finished. If this is not the case, please re-open.

PR #101, opened 2023-10-21
Fix/user openapi docs

CodeNamedRobin:
Fixed a minor issue where in the openapi-docs for the user-controller UserRequest was mentioned, but UpdateUserRequest was needed.


JustSamuel (approved):
Please fix merges before requesting review <3

PR #102, opened 2023-10-21
Added permissions for events to BAC role

CodeNamedRobin:
See title.


src/gewis/gewis.ts
```diff
@@ -356,6 +356,12 @@ export default class Gewis {
         Balance: {
           get: { all: star },
         },
+        Event : {
+          get: { all: star },
```
Yoronex, line 360: I think that `GET /events` and `GET /events/:id` should be put in the RBAC of general SudoSOS users, because also non-BAC people should be able to fill in events planning sheets. Think of feuten and BHV'ers. However, then everyone can fetch complete planning sheets, so that is not a good solution either. For now, I think this is fine. Maybe we can add a new role to be able to fetch and fill in planning sheets? @JustSamuel

src/gewis/gewis.ts
```diff
@@ -356,6 +356,12 @@ export default class Gewis {
         Balance: {
           get: { all: star },
         },
+        Event : {
```
Yoronex, line 359: Space after `:`?! How did the linter pass?

src/gewis/gewis.ts
```diff
@@ -356,6 +356,12 @@ export default class Gewis {
...
+          get: { all: star },
+          update: { all: star },
+          delete: { all: star },
+          create: { all: star },
+        },
```
Yoronex, line 364: You miss the "assign EventAnswer" rule. https://github.com/GEWIS/sudosos-backend/blob/e720d09bda53b81f9cdd2f9a28d43cd3cf740f29/src/controller/event-controller.ts#L89

Yoronex (changes requested):

JustSamuel:
Deprecated because of the new RBAC permissions setup.

Issue #21, opened 2023-03-01
Return HTTP 204 instead of 200 when returning an empty body.


GEWISabc:
**Description:**
Currently, our API handles DELETE requests for certain routes, such as DELETE /banners/{id}, by returning a response with an HTTP status code of 200 along with the deleted entity in the response body. However, according to HTTP standards, DELETE requests that have been successfully processed should return an HTTP status code of 204 with an empty response body.

**Task:**

- Modify the DELETE route handlers to return an HTTP status code 204 (No Content) upon successful deletion, signaling that the resource has been removed.
- Remove any unnecessary response data from the body of DELETE responses, ensuring they are empty as per the HTTP specification.
- Update relevant documentation or API descriptions to reflect this change in behavior.

**Expected Outcome:**
By addressing this issue, we aim to align our API responses with industry best practices and HTTP standards. This change will improve the consistency and correctness of our API, making it more intuitive for developers and clients consuming our services.

**Skills Required:**
- Backend development
- Familiarity with HTTP status codes and RESTful API design

**Difficulty Level:**
Beginner

**Additional Information:**
This issue is an excellent opportunity for contributors to participate in Hacktoberfest and make a meaningful contribution to our project. Feel free to ask any questions or seek guidance on how to implement the suggested changes. Your efforts will help improve the quality of our API and enhance the overall user experience.

JoHeyJo:
I PR a few changes. Is this what you had in mind? [https://github.com/GEWIS/sudosos-backend/pull/81](url)

JustSamuel:
Yeah! Something like that.

PR #133, opened 2024-01-07
Merge Develop into Main

JustSamuel:
- **Fixes HTTP status codes for DELETE and empty bodies (#129)**
  - Corrected HTTP return codes for DELETE requests and empty bodies (Issue #21).

- **Ensure that only LOCAL_USERS can get a reset password token (#127)**
  - Implemented restrictions to allow only LOCAL_USERS to retrieve reset password tokens.

- **Wraps stripe updates in a transaction (#130)**
  - Encapsulated stripe updates within a transaction for improved integrity.

- **Hotfix for stripe transaction erroring on deposit update (#131)**
  - Resolved a bug causing typeorm to error when saving deposit status within a transaction.

- **Hotfix stripe transaction bug (#132)**
  - Addressed a bug where typeorm would error when trying to save deposit status within a transaction.

### DIFF
--- a/src/controller/authentication-controller.ts
+++ b/src/controller/authentication-controller.ts
@@ -345,7 +345,7 @@ export default class AuthenticationController extends BaseController {
    * @operationId resetLocalWithToken
    * @tags authenticate - Operations of authentication controller
    * @param {AuthenticationResetTokenRequest} request.body.required - The reset token.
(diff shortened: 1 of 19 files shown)


rinkp:
Reminder to fast-forward `main`

Yoronex (approved):

Issue #108, opened 2023-11-22
Enforce minimum payment


rinkp:
> 3.4 Deposits
> When depositing money into their account, a user has to deposit exactly their total negative balance (such that their balance becomes 0 euros after the deposit) or at least 10 euros. The primary way for a deposit to be created is by online payment through the website. In exceptional cases, other ways to increase a balance (such as by invoice) can be agreed upon with the BAC.

Both frontend and backend lack this feature
@Gijsdeman

PR #135, opened 2024-01-11
Enforces a minimum top-up for stripe deposits.

JustSamuel:
Based on our TOS

### DIFF
--- a/src/service/stripe-service.ts
+++ b/src/service/stripe-service.ts
@@ -49,28 +49,14 @@ export default class StripeService {
   }
 
   /**
-   * Topup should be at least 10 euros or the users negative balance (or greater).
-   * Returns false if request amount is too low.
+   * Topup should be at least 10 euros or the users negative balance.
    * @param balance
    * @param request
    */
   public static validateStripeRequestAmount(balance: BalanceResponse, request: StripeRequest): boolean {
-    // User is in the negative and would remain negative.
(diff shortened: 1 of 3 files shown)


src/service/stripe-service.ts
```diff
@@ -46,6 +48,17 @@ export default class StripeService {
...
   }
 
+  /**
+   * Topup should be at least 10 euros or the users negative balance.
+   * @param balance
+   * @param request
+   */
+  public static validateStripeRequestAmount(balance: BalanceResponse, request: StripeRequest): boolean {
+    // Check if top up is enough
+    if (request.amount.amount > (process.env.MIN_TOPUP || 1000)) return true;
```
Yoronex, line 58: Shouldn't this be greater than or equal? Because now you cannot top up with 10 euros exactly if you don't have a debt.

src/service/stripe-service.ts
```diff
@@ -46,6 +48,17 @@ export default class StripeService {
...
+    return request.amount.amount === balance.amount.amount;
```
Yoronex, line 59: Isn't your balance negative and the to-be-deposited amount positive? Meaning that this will not work for negative balances? Still confused though, as there are test cases that seem to work?

Yoronex (approved):

rinkp:
Closes GH-108

PR #145, opened 2024-01-22
Fixed bug where bindings were not overwritten for the cron tasks.

JustSamuel:
The cron task runs in a different process than the main. This means that GEWIS bindings were never injected for the cron tasks and thus the LDAP sync created "normal" user accounts. This gave GEWIS users two accounts, one of their GEWIS web account and one for the LDAP sync.

### DIFF
--- a/src/cron.ts
+++ b/src/cron.ts
@@ -73,6 +73,9 @@ async function createCronTasks(): Promise<void> {
 
   application.tasks = [syncBalances];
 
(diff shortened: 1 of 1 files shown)


Kevinjil (approved):

Kevinjil (approved):
Even though the second commit doesn't cover the title of the PR, it indeed is a duplicate. LGTM.

PR #149, opened 2024-01-31
Feature/product ordering

### LINKED ISSUE #12: POSProductOrdering Controller (Ruben Brinkman)
Create a controller that manages product orderings for POS. It should contain a POS ID and and a list of product ID's in an ordering. It can be incomplete.

- [] Get POS Product Order by POS ID
- [] Patch POS Product Order by ID
- [] Post POS pPoduct Order by ID

SvenMokveldje:
I have finally implemented product ordering, not each product (revision) has 3 new booleans. Featured (which will be highlighted), preferred (e.g. borrel products) and showOnPriceList which is whether the products will be shown on the narrowcasting screens. All tests still pass but let me know if anything has to be changed.

Fixes #12

### DIFF
--- a/src/controller/request/product-request.ts
+++ b/src/controller/request/product-request.ts
@@ -24,9 +24,9 @@ export interface BaseProductParams {
   category: number;
   alcoholPercentage: number;
   name: string;
-  featured: boolean;
-  preferred: boolean;
-  showOnPriceList: boolean;
+  featured?: boolean;
+  preferred?: boolean;
+  priceList?: boolean;
 }
 
 export default interface CreateProductParams extends BaseProductParams {
@@ -46,9 +46,9 @@ export type ProductRequest = UpdateProductParams | CreateProductParams;
  * @property {number} vat.required - VAT group ID of the product
  * @property {number} category.required  - Category of the product
  * @property {number} alcoholPercentage.required  - Alcohol percentage of the product in 2 decimals
- * @property {boolean} featured.required - If product is featured
- * @property {boolean} preferred.required - If product is preferred
- * @property {boolean} showOnPriceList - If product is shown on narrowcasting screens
+ * @property {boolean} featured - If product is featured
+ * @property {boolean} preferred- If product is preferred
+ * @property {boolean} priceList - If product is shown on narrowcasting screens
  * @property {number} ownerId.required - ID of the owner
  */
 export interface CreateProductRequest extends BaseProductParams {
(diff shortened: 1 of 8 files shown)


src/controller/request/product-request.ts
```diff
@@ -43,6 +46,9 @@ export type ProductRequest = UpdateProductParams | CreateProductParams;
  * @property {number} vat.required - VAT group ID of the product
  * @property {number} category.required  - Category of the product
  * @property {number} alcoholPercentage.required  - Alcohol percentage of the product in 2 decimals
+ * @property {boolean} featured.required - If product is featured
```
JustSamuel, line 49: Why required? I would have them all be optional.

src/controller/request/product-request.ts
```diff
@@ -24,6 +24,9 @@ export interface BaseProductParams {
   category: number;
   alcoholPercentage: number;
   name: string;
+  featured: boolean;
+  preferred: boolean;
+  showOnPriceList: boolean;
```
JustSamuel, line 29: Maybe change to priceList to be more inline with the others?

src/controller/request/product-request.ts
```diff
@@ -56,5 +62,8 @@ export interface CreateProductRequest extends BaseProductParams {
  * @property {number} vat.required - VAT group ID of the product
  * @property {number} category.required  - Category of the product
  * @property {number} alcoholPercentage.required  - Alcohol percentage of the product in 2 decimals
+ * @property {boolean} featured.required - If product is featured
```
JustSamuel, line 65: again, I don't think you should have to specify this, but its fine since it will most likely be added to the FE as a checkmark anyhow.

src/controller/response/product-response.ts
```diff
@@ -44,6 +44,9 @@ export interface BaseProductResponse extends BaseResponse {
  *           excluding VAT
  * @property {string} image - The URL to the picture representing this product.
  * @property {number} alcoholPercentage.required - The percentage of alcohol in this product.
+ * @property {boolean} featured.required - If product is featured
+ * @property {boolean} preferred.required - If product is preferred
```
JustSamuel, line 48: JSDOC does not match interface.

JustSamuel (changes requested):
Small remarks but it looks good.

If you feel like it, you could add a test case where you filter on these properties.

src/controller/request/product-request.ts
CodeNamedRobin: Problem here is that UpdateProductParams and CreateProductParams are different. For creation, these params are required, for updating they are not.

src/controller/request/product-request.ts
CodeNamedRobin: I would just put this as false by default. FE implementation on creation doesn't really need to include this right? Just having a toggle on each product entry to feature it would be enough imo.

SvenMokveldje: I changed it such that they are always optional to fill in as indeed even at creation or updating you do not always need it.

Yoronex (approved):

JustSamuel (approved):
Lekker werk.

PR #151, opened 2024-02-05
Fix filtering and make it more consistent across the API and fix user debt notifications

### LINKED ISSUE #84: GET /balances/all cannot filter based on UserType
This field is ignored when supplying it with either integers or full names. Somehow, the test cases do seem to pass, so something is going wrong with parsing. Probably related with https://github.com/GEWIS/sudosos-backend/issues/77.

CodeNamedRobin:
Filtering is flawed in a couple places, and should be up to the latest spec (as seen on GET /balances/all). This PR should fix that, along with some filtering issues where the parameter seems to be simply ignored.

Fixes #84.

TODO:

- [x] Make invoice filtering actually work
- [x] Add "available values" to every filtering option with an enum, as on GET /users/

### DIFF
--- a/src/entity/user/user.ts
+++ b/src/entity/user/user.ts
@@ -53,6 +53,13 @@ export const TOSRequired = [
 ];
 
 /**
+ * All users that should be notified when in debt.
+ */
+export const NotifyDebtUserTypes: UserType[] = [
+  UserType.LOCAL_ADMIN, UserType.LOCAL_USER, UserType.MEMBER,
+];
+
+/**
  * @typedef {BaseEntity} User
  * @property {string} firstName.required - First name of the user.
  * @property {string} lastName - Last name of the user.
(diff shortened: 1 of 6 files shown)


src/entity/user/user.ts
```diff
@@ -52,6 +52,13 @@ export const TOSRequired = [
   UserType.MEMBER, UserType.LOCAL_USER, UserType.LOCAL_ADMIN,
 ];
 
+/**
+ * All users that should be notified when in debt.
+ */
+export const NotifyDebtUserTypes: UserType[] = [
```
JustSamuel, line 58: Not in the scope of this PR.

src/subscriber/transaction-subscriber.ts
```diff
@@ -61,6 +68,9 @@ export default class TransactionSubscriber implements EntitySubscriberInterface
     if (balanceBefore.amount.amount < 0) return;
     // User was not in debt before this new transaction
 
+    if (!(user.type in NotifyDebtUserTypes)) return;
```
JustSamuel, line 71: Note in the scope of this PR.

JustSamuel (approved):
Fine, but either expand the scope of this PR by changing the name or make another PR.

PR #162, opened 2024-04-16
feat: toggle extensive data processing

### LINKED ISSUE #161: Users cannot toggle extensive data processing
As per the ToS, users should be able to turn on/off extensive data processing. This is currently not possible.

JustSamuel:
Closes #161

### DIFF
--- a/test/unit/controller/user-controller.ts
+++ b/test/unit/controller/user-controller.ts
@@ -202,7 +202,7 @@ describe('UserController', (): void => {
         User: {
           get: own,
           acceptToS: own,
-          update: { own: new Set<string>(['firstName', 'lastName']) },
+          update: { own: new Set<string>(['firstName', 'lastName', 'extensiveDataProcessing']) },
         },
         Product: {
           get: own,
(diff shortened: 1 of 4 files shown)


test/unit/controller/user-controller.ts
```diff
@@ -906,6 +906,20 @@ describe('UserController', (): void => {
       expect(user.firstName).to.deep.equal(firstName);
       verifyUserResponse(spec, user);
     });
+    it('should allow user to update extensiveDataProcessing', async () => {
```
Yoronex, line 909: Maybe also add a sanity-check use case that checks that a) you cannot change any other properties of yourself when you are not an admin and b) you cannot change extensive data processing of other users? Because I'm a bit scared

Yoronex (approved):

test/unit/controller/user-controller.ts
JustSamuel: Doesn't really work in the test-env as we define our own rulesets for each token per test file. This means that you will only be testing if I wrote the correct test cases, but it won't actually test if the correct RBAC is defined in the production database.

I'll make an issue and approve this.

PR #169, opened 2024-04-22
bugfix: mysql test suite

### LINKED ISSUE #82: Add coverage test on MySQL/MariaDB instance to CI jobs
Currently, we only run the unit tests on an in-memory SQLite database. However, SQLite and MySQL/MariaDB differ in engine in syntax, so a query that works in one, might not work in the other. Unfortunately, we have some custom SQL queries (looking at you, balances) that use different clauses depending on whether SQLite or MariaDB is used. At the time of writing, the `develop` branch has a major defect, because the user cannot fetch its balance, which in turn is caused by the query not having the correct syntax. However, the CI passes just fine. To prevent such issues in the future, we should add a job to the CI that runs all unit tests on a MySQL/MariaDB instance.

Because such a database probably cannot be run/stored from memory, we will need to do some work ourselves in every test suite:
- Drop the complete database once a test suite has finished.
- Synchronize the database structure when starting a test suite (because we dropped it before).

### LINKED ISSUE #83: TypeORM synchronize failing on MySQL due to NFCAuthenticator
When trying to synchronize the database schema in a MySQL database instance, TypeORM throws an error about a duplicate index. This seems to be caused by the `NFCAuthenticator` entity. Other authenticators do somehow not have this issue, because they are instances of `HashAuthenticator`. Even though it might not provide many benefits, it would resolve this issue and make the codebase more consistent if `NFCAuthenticator` also extends from `HashAuthenticator`. Currently, NFC is not yet used in production, so we can safely change this entity.

### LINKED ISSUE #88: Add Stripe test keys to CI runner to also test Stripe integration in CI
Currently, Stripe is not tested in the coverage tests because its keys are missing. Therefore, the integration is not yet tested, but we should do that as not everyone tests their Stripe integration locally with their own keys.

JustSamuel:
The MySQL test suite now finally works. Some changes were needed such to never close a connection if we are in a MySQL context. 

The original mistake originated from the test/setup.ts file where the env vars were overwritten and always resulted in a sqlite context.

This PR also includes multiple bug fixes found during this process.

## **Changes**
To ensure future test cases also work make sure to use the following in your `before`
```ts
const connection = await Database.initialize();
await truncateAllTables(connection);
```

And the following in your after
```ts
await finishTestDB(ctx.connection);
```
This also means to speed up the testing, refrain from making this a `beforeEach`.

## TODO
- [x] MySQL: no user ids in the balance query crashes
- [x] Database tests have been commented out
- [x] A good migration testing setup

## Fixes & Closes
CI: Adds post migration test suite.
fix: `Invoice` migration, such that it follows the current schema synchronize.
fix: `product-revision.alcoholPercentage` had precision: 0.
fix: bug in balance service that would crash mysql on empty ids array.
fix: `FineHandoutEvents` were not sorted by default
fix: `EventService` filtering on `beforeDate` was wrong
fix: `InvoiceService` latest valid invoice function was wrong
fix: MySQL didn't store milliseconds for timestamps.
fix: seed script for MySQL.

Closes #82
Closes #83
Closes #88


test/setup.ts
```diff
@@ -36,11 +37,20 @@ use(sinonChai);
 use(chaiSorted);
 use(deepEqualInAnyOrder);
 
+config();
 process.env.NODE_ENV = 'test';
-process.env.HTTP_PORT = '3001';
-process.env.TYPEORM_CONNECTION = 'sqlite';
-process.env.TYPEORM_DATABASE = ':memory:';
-process.env.TYPEORM_SYNCHRONIZE = 'true';
+if (!process.env.TYPEORM_CONNECTION) {
```
Yoronex, line 42: This execution flow needs to be tweaked somehow. Because in a development setting, you probably have your own "production" database in a SQLite file while you still write your test cases. You want to run these test cases 90% of the time from an in-memory SQLite database, but only use an actual database file when you want to debug a test case. I think the database error when running a test locally is very confusing, because it is not entirely clear what database is being used.

src/service/balance-service.ts
```diff
@@ -184,6 +184,15 @@ export default class BalanceService {
   public static async getBalances({
     ids, date, minBalance, maxBalance, hasFine, minFine, maxFine, userTypes, orderDirection, orderBy,
   }: GetBalanceParameters, pagination: PaginationParameters = {}): Promise<PaginatedBalanceResponse> {
+    // Return the empty response if request has no ids.
+    if (ids?.length === 0) {
+      const { take, skip } = pagination;
+      return {
+        _pagination: { take, skip, count: 0},
```
Yoronex, line 191: This line has a linting error, but the linter seems to pass in the Github runner. Something is going wrong there...

test/unit/service/debtor-service.ts
```diff
@@ -335,14 +335,28 @@ describe('DebtorService', (): void => {
...
+        await queryRunner.release();
+      }
     }
 
     before(async () => {
       await clearFines();
-      const fineTransfers = (await Transfer.find({ relations: ['fine'] })).filter((t) => t.fine != null);
-      expect(fineTransfers.length).to.equal(0);
+      // const fineTransfers = (await Transfer.find({ relations: ['fine'] })).filter((t) => t.fine != null);
+      // expect(fineTransfers.length).to.equal(0);
```
Yoronex, line 359: Why are these lines commented?

test/unit/service/debtor-service.ts
```diff
@@ -335,14 +335,28 @@ describe('DebtorService', (): void => {
 
       await Fine.clear();
       await Transfer.remove(fineTransfers);
-      await FineHandoutEvent.clear();
-      await UserFineGroup.clear();
+
+      // Needed for non sqlite databases.
```
Yoronex, line 339: Why is this needed? I do not understand why the FineHandoutEvent and UserFineGroup tables should be emptied after seeding.

test/unit/service/event-service.ts
```diff
@@ -274,11 +275,11 @@ describe('eventService', () => {
         startDate: event.startDate,
       });
     });
-    it('should send a reminder to events starting in two days from now', async () => {
+    it('should send a reminder to events starting in two days and an hour from now', async () => {
       const event = ctx.events[0];
       const now = new Date();
       const users = Array.from(new Set(event.answers.filter((a) => a.availability == null).map((a) => a.user)));
-      const startDate = new Date(now.getTime() + 1000 * 3600 * 24 * 2);
+      const startDate = new Date(now.getTime() + 1000 * 3600 * 24 * 2 + 1000);
```
Yoronex, line 282: This calculation adds 2 days and 1 second, not 2 days and 1 hour.

Yoronex (changes requested):

test/unit/service/debtor-service.ts
JustSamuel: thats your code

src/database/database.ts
```diff
@@ -148,9 +152,52 @@ const options: DataSourceOptions = {
...
+      else return Promise.resolve(createConnection(options));
+
+    } else if (process.env.TYPEORM_CONNECTION === 'sqlite') {
+      // And for sqlite we always just create a new connection.
+      return Promise.resolve(createConnection(options));
+    } else {
+      throw new Error(`Unsupported connection type ${process.env.TYPEORM_CONNECTION}`);
+    }
+  },
+  finish: async (connection: DataSource) => {
```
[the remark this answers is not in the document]
Yoronex: That is a reasonable concern. I think the second option is then a good alternative. Or move this finish function to the test folder, because it seems to be only used by test suites

test/setup.ts
```diff
@@ -74,3 +83,42 @@ export async function generateKeys(): Promise<{ publicKey: string, privateKey: s
 export function sourceFile(file: string) {
   return file.replace('out/test/', 'test/').replace('.js', '.ts');
 }
+
+export const PERSISTENT_TEST_DATABASES = new Set(['mysql', 'mariadb']);
```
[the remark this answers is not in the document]
Yoronex: You marked this as resolved but the duplication is still here? Why?

test/setup.ts
```diff
@@ -36,11 +37,20 @@ use(sinonChai);
...
+  console.log('Setting sqlite defaults');
+  process.env.HTTP_PORT = '3001';
+  process.env.TYPEORM_CONNECTION = 'sqlite';
+  process.env.TYPEORM_DATABASE = ':memory:';
+  process.env.TYPEORM_SYNCHRONIZE = 'true';
+}
+
+// Always sync in sqlite.
+if (process.env.TYPEORM_CONNECTION === 'sqlite') {
+  process.env.TYPEORM_SYNCHRONIZE = 'true';
```
Yoronex, line 52: If we do not wish to override SQLite defaults, why do we then still override this particular variable? I can assume there are use cases when you want to use a local database file, but do not want it to synchronize (for example when the synchronization is doing something wrong?).

test/unit/service/event-service.ts
```diff
@@ -274,11 +275,11 @@ describe('eventService', () => {
         startDate: event.startDate,
       });
     });
-    it('should send a reminder to events starting in two days from now', async () => {
+    it('should send a reminder to events starting in two days and an second from now', async () => {
```
Yoronex, line 278: _a_ second ;)

test/unit/service/event-service.ts
```diff
@@ -310,10 +311,10 @@ describe('eventService', () => {
         startDate: event.startDate,
       });
     });
-    it('should not send a reminder to events starting in three days and one millisecond from now', async () => {
+    it('should not send a reminder to events starting in three days and one minute from now', async () => {
```
Yoronex, line 314: The timestamp you create below is 4 days and one second from now, not three days and one minute

Yoronex (changes requested):

test/unit/service/event-service.ts
Yoronex: Still a second, not a minute :\

Yoronex (approved):
Still a small typo you can easily fix, but the rest is nice.

PR #179, opened 2024-05-20
feat: maximum top-up amount check

tomudding:
Introduces validation for the maximum amount a top-up can have; to verify that the user can only top-up their balance with €150 minus their current balance.

### DIFF
--- a/src/controller/stripe-controller.ts
+++ b/src/controller/stripe-controller.ts
@@ -74,11 +74,18 @@ export default class StripeController extends BaseController {
     try {
       const amount = Dinero({ ...request.amount } as DineroObject);
       const balance = await BalanceService.getBalance(req.token.user.id);
-      if (!StripeService.validateStripeRequestAmount(balance, request)) {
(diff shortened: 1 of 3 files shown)


tomudding:
Based on internal feedback the code has been changed to remove the "fix" that would only allow you to top up with exactly your negative balance. Original code can be found at https://github.com/GEWIS/sudosos-backend/compare/develop...tomudding:sudosos-backend:feature/maximum-topup-check-original.

If the original discussion had been documented/mentioned in GH-135 this could have been prevented. So for future reference, the discussion resulted in determining that the wording is ambiguous, however, the behaviour from SuSOS should be kept and thus the wording in the current version of the TOS* is an actual OR and it does not have an implied `currentBalance >= 0` for the second statement.

Current version of the TOS: <blockquote>exactly their total negative balance (such that their balance becomes 0 euros after the deposit) or at least 10 euros </blockquote>

JustSamuel (approved):
Mooi

Issue #180, opened 2024-05-29
`StripeController` test failure due to too much balance


tomudding:
After the merge of GH-179 one specific test related to Stripe deposists keeps failing:

https://github.com/GEWIS/sudosos-backend/blob/cdcfb6dec7fd86641fd26bb5e157acee054be291/test/unit/controller/stripe-controller.ts#L154

Analysis shows that the [`seedStripeDeposits`](https://github.com/GEWIS/sudosos-backend/blob/cdcfb6dec7fd86641fd26bb5e157acee054be291/test/seed.ts#L1001) can/will create 3 valid Stripe deposits resulting in a balance of 117 euros. A top-up of another 39 euros will thus fail as this exceeds the 150 euro balance limit.

PR #181, opened 2024-05-29
fix: `StripeController` test with too high test balance

tomudding:
Analysis shows that the `seedStripeDeposits` can/will create 3 valid Stripe deposits resulting in a balance of 117 euros. A top-up of another 39 euros will thus fail as this exceeds the 150 euro balance limit.

Instead of changing the seeding (which may inadvertantly affect other tests) we simply change the amount of the top-up to €19.50.

Fixes GH-180.


tomudding:
@JustSamuel if necessary I can also add two other tests to check the minimum and maximum tests.

JustSamuel (approved):
Nice, thanks!

Issue #157, opened 2024-02-17
Useful that all transactions/transfers endpoints return PaginatedFinancialMutationResponse


SuperVK:
Currently all certain endpoints such as users/{id}/transfers or users/{id}/transactions currently return PaginatedTransferResponse and PaginatedBaseTransactionResponse. However these are usually all handled in the same way by the frontend. So even though it seems a bit weird because it is already guaranteed that the type of each transaction/transfer is one or the other, it is easier to change these all to PaginatedFinancialMutation. This way calls from both users/{id}/transfer and pointofsale/{id}/transactions can all be handled as the same type, and be rendered in the exact same component without weird type guarding.

JustSamuel:
@Yoronex opinion?

Yoronex:
I prefer not to do this, as the PaginatedFinancialMutation response adds a lot (and I mean _a lot_ to the separate endpoints). I see that it is easier for the frontend to handle as you no longer need to map all these objects there, but I am not convinced that it outweighs the extra bloat that we have to add to those respective endpoints.

SuperVK:
oh I thought i already closed this, I solved this in another way that works much better :)

PR #183, opened 2024-05-31
Remove createdBy transactions on the user related endpoints.

SuperVK:
Remove transactions created by the user from the transactions overview of the user (as well as the financial mutation)

# Description
Currently, financial mutations and transactions on `/user/{id}/transactions` and `/user/{id}/financialmutations` return also transactions that are created by the user. This PR removes that functionality. The idea is that all the financial mutation (transactions, transfers, and financialmutation) on the `/user/*` should only return transactions/transfers that mutate the user's balance. 

This is also tied into the frontend, where it would only make sense to display only transactions that influence the balance of the user together, serving the purpose of being able to see why you are €-15 when you topped up a week ago. Transactions that are created by the user should be displayed in a separate view because it serves a different purpose (namely, looking back at who you bought for in case someone comes knocking after an event). 

This also solves (not fixes) the need for #157, because now transactions/transfers from users will actually be treated in a different way than transactions from an endpoint. They will get different views based on their purpose (seeing influences or your balance, or administratively what transactions happened on a PoS or by a user).

## Types of changes

### DIFF
--- a/src/controller/user-controller.ts
+++ b/src/controller/user-controller.ts
@@ -1001,7 +1001,7 @@ export default class UserController extends BaseController {
 
   /**
    * GET /users/{id}/transactions
-   * @summary Get an user's transactions (from, to or created)
+   * @summary Get transactions from a user.
    * @operationId getUsersTransactions
    * @tags users - Operations of user controller
    * @param {integer} id.path.required - The id of the user that should be involved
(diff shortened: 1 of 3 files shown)


JustSamuel:
And you suggest instead using a `GET transactions` to find which transactions a user created that were not for him? iirc you cant filter the user out so that would have duplicates?

Or do you think its not needed anymore that a user can see which transactions were created in his name for someone else?

JustSamuel (approved):

SuperVK:
Exactly this point was a problem in my head as well. I do think being to see which transactions you created is needed. I think a filter per POS will solve, because somebody from the BAC or from BOOM can filter on their respective POS and mostly see transactions made for others, rarely for themselves. However if you filter on "BAC Verkoop" you will still a lot of yourself, so i am not sure about that.

JustSamuel:
You can merge it imho, but if you need a follow up, please create an issue now.

PR #186, opened 2024-06-19
refactor: Make generating/calculating invoices and involving substructures separate from saving

CodeNamedRobin:
# Description
This PR is mostly preparing for an eventual "invoice preview" option. Before it was the case that both calculating and saving to the database was in the same function each time, these are now split.
None

## Types of changes
Refactor


JustSamuel:
Already included in https://github.com/GEWIS/sudosos-backend/pull/259

Issue #155, opened 2024-02-15
[Feature Request] Endpoint to generate a fine overview


CodeNamedRobin:
As the BAC treasurer needs to book in the fines (and would probably do this monthly), it would be good to have a way to get an overview of the amount of fines that we're dished out in a certain time period. This would in my opinion be very similar to the current transaction report feature, although that is for all users, and this feature is specific to the BAC treasurer.

JustSamuel:
Endpoint for the report in the form of a json is in PR #189, pdf will have to wait a for a bit.

JustSamuel:
PDF generation is currently live, re-open if you are missing something.

PR #189, opened 2024-07-01
Feature: fine reporting

JustSamuel:
# Description
Route for getting an overview of all the handed out and waived fines. Next step will be allowing it to be reported as a pdf.

#155 (does not close)

## Types of changes

### DIFF
--- a/test/unit/controller/debtor-controller.ts
+++ b/test/unit/controller/debtor-controller.ts
@@ -568,6 +568,20 @@ describe('DebtorController', () => {
       expect(report.toDate).to.equal(toDate.toISOString());
     });
 
+    it('should return correct model', async () => {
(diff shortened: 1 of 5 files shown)


test/unit/controller/debtor-controller.ts
```diff
@@ -548,4 +548,81 @@ describe('DebtorController', () => {
       expect(res.status).to.equal(400);
     });
   });
+
+  describe('GET /fines/report', () => {
```
Yoronex, line 552: Maybe also add a test case for your sanity check (that a transfer cannot have a corresponding fine and waivedFine)? Increases test coverage and a higher number is always better.

Yoronex (changes requested):

Yoronex (approved):

Issue #167, opened 2024-04-21
Fines can be handed out to suspended/deleted accounts


Yoronex:
Users that are marked as `deleted`, can be fined. This has already happened with production user `1907`. This seems to be caused by the fact that deleted users are included in the `GET /balances/all` response, but this probably should not be the case. Their balance is however cached, which I think is fine?

To fix, deleted users should not be included in the `BalanceService.getBalances` repsonse, as this method is also included in the execution pipeline to calculate and hand out fines. Seems to be an easy fix (a simple WHERE clause). However, the BAC treasurer should also check which deleted people received a fine and waive these, because they have no way to pay their fine.

Yoronex:
@CodeNamedRobin

PR #190, opened 2024-07-01
Fix: balance deleted users

JustSamuel:
# Description
Adds query param for if deleted users should be returned by the balance query

#167 

## Types of changes


src/service/balance-service.ts
```diff
@@ -305,12 +309,12 @@ export default class BalanceService {
     if (skip) recordsQuery += `offset ${skip} `;
 
     const balances = await connection.query(recordsQuery, parameters);
-
     if (balances.length > 0 && balances[0].amount === undefined) {
       throw new Error('No balance returned');
     }
+    let count = (await connection.query(query, parameters)).length;
```
Yoronex, line 315: Waarom is deze `const` nu een `let` geworden? Hij lijkt namelijk nergens anders geredefined te worden?

Yoronex (approved):

PR #192, opened 2024-07-01
fix: retry test case to fix flaky test

JustSamuel:
# Description
Test cases has a tendency to fail.

## Types of changes

### DIFF
--- a/test/unit/service/event-service.ts
+++ b/test/unit/service/event-service.ts
@@ -901,7 +901,8 @@ describe('eventService', () => {
   });
 
   describe('deleteEvent', () => {
-    it('should correctly delete an event with its answer sheets', async () => {
+    it('should correctly delete an event with its answer sheets', async function () {
+      this.retries(3);
       const event = ctx.events[0];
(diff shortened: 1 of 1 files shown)


Yoronex:
Any idea why the test case sometimes fails?

Yoronex:
Can you send me more details, for example where it fails and with what reason? I have never seen this test case fail.

Yoronex:
Okay I have encountered the issue myself.

image

This seems to happen only in the `test-ci-migrate` tests, which use the database after migrations. The issue is that event answer sheets are not correctly (automatically) deleted by the database, even though they have the SQL relational property `CASCADE`. Any clue what the issue could be? Database still deleting these sheets in the background?

JustSamuel:
I don't think its the `migrate`, see the current dev flow:
image

Maybe you are just to fast? I.E the cascade delete taking longer than the next step to start?

Yoronex:
Not the only test that seems to be flaky...

image

JustSamuel:
Seems to have fixed itself over the course of the last month, closing for now, can be re-opened if found to be an issue again.

PR #201, opened 2024-07-03
Refactor(container-service): moved back to TypeORM FindWhere instead of custom query building

JustSamuel:
# Description
Long overdue, will most likely hit a lot of conflicts (@Yoronex). But I will try and keep it small PRs (so POS, and product separate)

## Related issues/external references
#153 

## Types of changes
- Bug fix _(non-breaking change which fixes an issue)_
- New feature _(non-breaking change which adds functionality)_

Issue #153, opened 2024-02-07
Refactoring service classes to improve cross service usability.


JustSamuel:
[In this pull request](https://github.com/GEWIS/sudosos-backend/pull/148) I also refactored the service class, since it was long over due.

However, given that I wrote quite a lot of the current service classes they mostly have the same issues. Service files / classes **should** only consider database entities, it is up to the controller to turn this entity into a response.

This prevents the issue I had where you go from service -> response  -> back to an entity. Wasting database queries and muddling up the code.

A following issue that I found is that there is a sneaky form of code duplication. Namely that we use `User.find()` in quite a lot of places, and do not use the UserService. Which is fine, however, lets say we want to implement [account expiry](https://github.com/GEWIS/sudosos-backend/issues/142) we would have to refactor a lot of places just to add something along the lines of `expired: false` to the find options. 

Another code smell this solves is the `.records` being used because most service functions return a paginated response, meaning we are wasting computing power on creating pagination just to throw it away.

To solve this I propose the following:

- Service classes **should** never return `Response`, but entities instead.
- Each entity  should have a service class with a `getOptions` function handles creating the TypeOrm `FindManyOptions` for this entity. 
(This would solve the User issue by turning the `User.find()` into a `User.find(UserService.getOptions())` , which would contain defaults regarding delation, expiry etc.)
- Controllers are responsible for turning entities into responses.

Yoronex:
> Another code smell this solves is the .records being used because most service functions return a paginated response, meaning we are wasting computing power on creating pagination just to throw it away.

This is not true, because when you do not give any pagination parameters, pagination is not actually applied.

JustSamuel:
> This is not true, because when you do not give any pagination parameters, pagination is not actually applied.

I am still of the opinion that `.records` is smelly, `.records[0]` is also super ugly.

Yoronex:
Then maybe as a middleground, we can implement a response like TypeORM does with `.findAndCount()`. I think that looks quite neat.

JustSamuel:
Note that this refactor per service should also include moving to a class "manager"  and away from the global TypeORM manager.

PR #202, opened 2024-07-07
Refactor: product service

JustSamuel:
NB: This should be merged after #201 

# Description
This PR, similar to PR #201, moves the query building back to findWhereOptions. This does have some API changes, namely the ability to filter on pos or container when getting products is removed. However, once I looked like it, that function was never exposed to the api...

Anyhow this should make the code significantly more maintainable 

#153 

## Types of changes


Yoronex:
Reviewed https://github.com/GEWIS/sudosos-backend/pull/203

PR #203, opened 2024-07-07
Refactor: point of sale service

JustSamuel:
# Description
Same story as the container and product service.

#153 

## Types of changes

### DIFF
--- a/src/service/point-of-sale-service.ts
+++ b/src/service/point-of-sale-service.ts
@@ -218,6 +218,14 @@ export default class PointOfSaleService {
       container: {
         owner: true,
       },
+      products: {
+        product: {
+          image: true,
+          owner: true,
+        },
+        category: true,
+        vat: true,
+      },
     };
 
     let revisionFilter: any = {};
--- a/src/service/point-of-sale-service.ts
+++ b/src/service/point-of-sale-service.ts
@@ -39,7 +39,6 @@ import {
 } from '../controller/request/point-of-sale-request';
 import { parseUserToBaseResponse } from '../helpers/revision-to-response';
 import AuthenticationService from './authentication-service';
-import { AppDataSource } from '../database/database';
 
 /**
  * Define point of sale filtering parameters used to filter query results.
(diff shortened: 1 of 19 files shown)


src/service/product-service.ts
```diff
@@ -154,85 +135,63 @@ export function parseGetProductFilters(req: RequestWithToken): ProductFilterPara
...
-      vat,
+      vat: {
+        id: revision.vat.id,
+        percentage: revision.vat.percentage,
+        hidden: revision.vat.hidden,
+      },
     };
   }
 
+  private static  revisionSubQuery(revision?: number): string {
```
Yoronex, line 178: Two spaces :(

test/unit/service/point-of-sale-service.ts
```diff
@@ -252,7 +252,7 @@ describe('PointOfSaleService', async (): Promise<void> => {
         useAuthentication: true,
       };
 
-      const response = await PointOfSaleService.directPointOfSaleUpdate(update);
+      const response = (await PointOfSaleService.updatePointOfSale(update)) as PointOfSaleWithContainersResponse;
```
Yoronex, line 255: I think the update function should have a default response (with or without containers/products) such that the responses are also always consistent.

Yoronex (changes requested):

src/service/product-service.ts
```diff
@@ -120,21 +106,16 @@ export interface ProductFilterParameters {
   priceList?: boolean;
 }
 
+// TODO figure out why this is unused
```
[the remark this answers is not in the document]
JustSamuel: Nah, it is just never used, as in: nowhere does a controller expose the ability to use the filter params. However fixing this is outside the scope of this PR. 

I have updated the TODO.

src/service/point-of-sale-service.ts
```diff
@@ -163,50 +128,23 @@ export default class PointOfSaleService {
...
-    return {
-      name: pointOfSale.name,
-      owner: parseUserToBaseResponse(pointOfSale.pointOfSale.owner, false),
-      id: pointOfSale.pointOfSale.id,
-      revision: pointOfSale.revision,
-    } as PointOfSaleResponse;
-  }
-
-  public static async applyPointOfSaleUpdate(base: PointOfSale, update: BasePointOfSaleParams) {
+  public static async updatePointOfSale(update: UpdatePointOfSaleParams) {
```
Yoronex, line 146: This function still has an `|` in its return type (`PointOfSaleResponse` or `PointOfSaleWithContainersResponse`). Please make explicit which type is returned, as this does not seem to be a choice for the user.

Yoronex (approved):

Issue #20, opened 2023-03-01
[Feature] Add soft deleting product / container / POS


GEWISabc:
https://gitlab.com/sudosos/sudosos-back-end/-/issues/56

PR #205, opened 2024-07-07
Feature: add deleting products, containers and points of sale

Yoronex:
Adds endpoints for deleting products, containers and points of sale.

# Description

The three aforementioned entities could not be deleted from the database, because that would invalidate the data, as transactions have a reference to their point of sale, container and product. To solve this, products, containers and points of sale are soft deleted under the hood. When they are soft deleted, they will never be returned in any endpoint except in the context of transactions. Soft deleted products, containers and points of sale can also not be used as correlated entities or within transactions.

When soft deleting, the entity corresponding entity is only marked as soft deleted. However, due to the aforementioned restrictions, it will never be shown again (even though a soft deleted product might still be included in a container). However, when a container is updated, it cannot be included any more.

Note that the seeder now also includes these soft deleted products, containers and points of sale. 

https://github.com/GEWIS/sudosos-backend/issues/20

## Types of changes

### DIFF
--- a/src/controller/container-controller.ts
+++ b/src/controller/container-controller.ts
@@ -36,6 +36,7 @@ import {
   UpdateContainerRequest,
 } from './request/container-request';
 import userTokenInOrgan from '../helpers/token-helper';
+import { IsNull, Not } from 'typeorm';
 
 export default class ContainerController extends BaseController {
   private logger: Logger = log4js.getLogger('ContainerController');
@@ -141,7 +142,7 @@ export default class ContainerController extends BaseController {
     // Handle request
     try {
       // Check if we should return a 404.
-      const exist = await ContainerRevision.findOne({ where: { container: { id: containerId } } });
+      const exist = await ContainerRevision.findOne({ where: { container: { id: containerId, deletedAt: IsNull() } } });
       if (!exist) {
         res.status(404).json('Container not found.');
         return;
--- a/src/controller/container-controller.ts
+++ b/src/controller/container-controller.ts
@@ -36,7 +36,7 @@ import {
   UpdateContainerRequest,
 } from './request/container-request';
 import userTokenInOrgan from '../helpers/token-helper';
-import { IsNull, Not } from 'typeorm';
+import { IsNull } from 'typeorm';
 
 export default class ContainerController extends BaseController {
   private logger: Logger = log4js.getLogger('ContainerController');
@@ -75,6 +75,10 @@ export default class ContainerController extends BaseController {
           body: { modelName: 'UpdateContainerRequest' },
           policy: async (req) => this.roleManager.can(req.token.roles, 'update', await ContainerController.getRelation(req), 'Container', ['*']),
           handler: this.updateContainer.bind(this),
+        },
+        DELETE: {
+          policy: async (req) => this.roleManager.can(req.token.roles, 'delete', await ContainerController.getRelation(req), 'Container', ['*']),
+          handler: this.deleteContainer.bind(this),
         },
       },
       '/:id(\\d+)/products': {
@@ -297,6 +301,40 @@ export default class ContainerController extends BaseController {
       res.json(await ContainerService.updateContainer(request));
     } catch (error) {
       this.logger.error('Could not update container:', error);
+      res.status(500).json('Internal server error.');
+    }
+  }
+
+  /**
+   * DELETE /containers/{id}
+   * @summary (Soft) delete the given container. Cannot be undone.
+   * @operationId deleteContainer
+   * @tags containers - Operations of container controller
+   * @param {integer} id.path.required - The id of the container which should be deleted
+   * @security JWT
+   * @return {string} 204 - Success
+   * @return {string} 404 - Not found error
+   * @return {string} 500 - Internal server error
+   */
+  public async deleteContainer(req: RequestWithToken, res: Response): Promise<void> {
+    const { id } = req.params;
+    this.logger.trace('Delete container', id, 'by user', req.token.user);
+
+    try {
+      const containerId = parseInt(id, 10);
+
+      const container = await Container.findOne({ where: { id: containerId } });
+
+      if (container == null) {
+        res.status(404).json('Container not found');
+        return;
+      }
+
+      await ContainerService.deleteContainer(containerId);
+      res.status(204).send();
+      return;
+    } catch (error) {
+      this.logger.error('Could not delete container', error);
       res.status(500).json('Internal server error.');
     }
   }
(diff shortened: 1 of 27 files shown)


Yoronex:
@JustSamuel do you perhaps have any idea why an `InvoiceService` test case is suddenly failing due to a type error during runtime? I have not changed anything around invoices?

JustSamuel (changes requested):

JustSamuel:
In general, I think this should still work with the refactor, but now you would simply add a param to the getOptions of each service where you specify if deleted are allowed or not.

src/service/product-service.ts
```diff
@@ -457,15 +457,26 @@ export default class ProductService {
...
+      };
+      await ContainerService.directContainerUpdate(update);
     }
   }
+
+  /**
+   * (Soft) delete a product
+   * @param productId
+   */
+  public static async deleteProduct(productId: number): Promise<void> {
```
JustSamuel, line 475: Doesn't it make more sense that this triggers a update propegation? If I delete a product it should also be removed out of all containers and POS right?

src/service/invoice-pdf-service.ts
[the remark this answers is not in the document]
Yoronex: Indeed out of scope, but this PR somehow broke this test case. I still have no idea why, but without it the CI fails. Do you want it in a separate PR?

src/service/product-service.ts
Yoronex: I also thought about it some more yesterday and today. By definition of the soft deletions, it should not be necessary (as deleted products should never, ever be returned again). It also reduces the amounts of "states" you can have, because if you always propagate after a soft delete, what happens if - somehow - you don't?

I'll keep these tests in place to verify that deleted products/containers/POSs are never returned again, but I will add the propagation. This is more to meet possible future use cases, like the one you mentioned over Whatsapp: changing transactions and wanting to know what products where in that specific container. It is possible with the current implementation, but it is slightly more complex.

src/controller/container-controller.ts
```diff
@@ -304,6 +309,40 @@ export default class ContainerController extends BaseController {
...
+   * @return {string} 500 - Internal server error
+   */
+  public async deleteContainer(req: RequestWithToken, res: Response): Promise<void> {
+    const { id } = req.params;
+    this.logger.trace('Delete container', id, 'by user', req.token.user);
+
+    try {
+      const containerId = parseInt(id, 10);
+
+      const container = await Container.findOne({ where: { id: containerId } });
```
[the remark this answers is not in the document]
Yoronex: Not necessary due to the implementation of soft deletes in TypeORM.

src/controller/product-controller.ts
```diff
@@ -291,6 +295,40 @@ export default class ProductController extends BaseController {
...
+   * @return {string} 500 - Internal server error
+   */
+  public async deleteProduct(req: RequestWithToken, res: Response): Promise<void> {
+    const { id } = req.params;
+    this.logger.trace('Delete product', id, 'by user', req.token.user);
+
+    try {
+      const productId = parseInt(id, 10);
+
+      const product = await Product.findOne({ where: { id: productId } });
```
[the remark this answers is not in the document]
Yoronex: Not necessary due to the implementation of soft deletes in TypeORM.

src/service/invoice-pdf-service.ts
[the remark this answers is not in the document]
JustSamuel: No its fine, given that it fixes CI.

src/controller/container-controller.ts
```diff
@@ -304,6 +309,40 @@ export default class ContainerController extends BaseController {
     }
   }
 
+  /**
+   * DELETE /containers/{id}
+   * @summary (Soft) delete the given container. Cannot be undone.
```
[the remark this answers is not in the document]
JustSamuel: Agree

src/controller/point-of-sale-controller.ts
```diff
@@ -361,6 +365,40 @@ export default class PointOfSaleController extends BaseController {
...
+   * @return {string} 500 - Internal server error
+   */
+  public async deletePointOfSale(req: RequestWithToken, res: Response): Promise<void> {
+    const { id } = req.params;
+    this.logger.trace('Delete point of sale', id, 'by user', req.token.user);
+
+    try {
+      const pointOfSaleId = parseInt(id, 10);
+
+      const pointOfSale = await PointOfSale.findOne({ where: { id: pointOfSaleId } });
```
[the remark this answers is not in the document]
Yoronex: Not necessary due to the implementation of soft deletes in TypeORM.

src/controller/container-controller.ts
```diff
@@ -140,15 +144,12 @@ export default class ContainerController extends BaseController {
 
     // Handle request
     try {
-      // Check if we should return a 404.
-      const exist = await ContainerRevision.findOne({ where: { container: { id: containerId } } });
-      if (!exist) {
+      const container = (await ContainerService
```
JustSamuel, line 147: Isn't 
`await ContainerRevision.findOne({...ContainerService.getOptions({containerId, returnProducts})})` cleaner? since you force it to be a findOne.

src/service/product-service.ts
```diff
@@ -326,12 +362,17 @@ export default class ProductService {
       category: true,
     };
 
-    const userFilter: any = {};
+    if (params.returnContainers) relations.containers = {
```
JustSamuel, line 365: Do products even have a link to containers?

JustSamuel (approved):
Already checked most of it, so just one question.

src/service/product-service.ts
Yoronex: I created it. It was missing, but adding it made the implementation of the `product-service`, `container-service` and `point-of-sale-service` more consistent.

src/controller/container-controller.ts
Yoronex: But I thought we wanted to be less dependent on different query implementations everywhere? The only thing that `findOne` does compared to `find`, is that `findOne()` basically does `find()[0]`. It is still undefined. By directly calling the service, we only query the entity once (and correctly).

JustSamuel: I thought it was more about setting "defaults". We can hide defaults within the getOptions (maybe an `archived` property later?)

Yoronex: But then what is the difference between using the container service or manually querying a revision with default options? Because the service method does the same thing

JustSamuel: Not a lot, I just think that its cleaner to "drop" the records[0], imho.

JustSamuel: (You can resolve this btw, since I am neutral about it given what you said)

PR #206, opened 2024-07-08
Feature: Payout Request PDF

JustSamuel:
# Description
This PR generalizes the concept of `entity -> pdf` by allowing a entity to be `Pdfable`. This means that the entity has a property `pdf` and implements `getPdfParamHash()` which should return a string that represent the hash of all properties that influence the generation of the pdf.

Or more specific, the `getPdfParamHash` should only change if the corresponding pdf of the property also changed. Moreover, the `Pdfable` should implement `createPdf` which returns a `Pdf` of the correct type.

## Types of changes


src/controller/invoice-controller.ts
```diff
@@ -323,13 +323,15 @@ export default class InvoiceController extends BaseController {
     this.logger.trace('Get Invoice PDF', id, 'by user', req.token.user);
 
     try {
-      const invoice = await InvoicePdfService.getOrCreatePDF(invoiceId);
+      const invoice = await Invoice.findOne({ ...InvoiceService.getOptions({ invoiceId }) });
```
Yoronex, line 326: Why use object spreading when you are not merging any objects?

src/entity/transactions/payout-request.ts
```diff
@@ -26,8 +26,12 @@ import Transfer from './transfer';
 import DineroTransformer from '../transformer/dinero-transformer';
 // eslint-disable-next-line import/no-cycle
 import PayoutRequestStatus from './payout-request-status';
+import PayoutRequestPdf from '../file/payout-request-pdf';
+import { hashJSON } from '../../helpers/hash';
+import PayoutRequestPdfService from '../../service/payout-request-pdf-service';
 
 @Entity()
+// TODO Migration
```
Yoronex, line 34: TODO

src/service/payout-request-pdf-service.ts
```diff
@@ -0,0 +1,108 @@
...
+   * @param payoutRequest - The payout request for which to generate the parameters.
+   * @returns {PayoutParameters} An instance of `PayoutParameters` containing all necessary information for PDF generation.
+   */
+  static getParameters(payoutRequest: PayoutRequest): PayoutParameters {
+    return new PayoutParameters({
+      payout: new Payout({
+        bankAccountName: payoutRequest.bankAccountName,
+        bankAccountNumber: payoutRequest.bankAccountNumber,
+        amount: payoutRequest.amount.getAmount(),
+        reference: `SDS-PR-${payoutRequest.id}`,
```
Yoronex, line 55: Maybe prefix the ID with some zeroes, so the number always has a length of 3 or 4 digits. `SDS-PR-001` looks nicer than `SDS-PR-1`.

src/service/payout-request-pdf-service.ts
```diff
@@ -0,0 +1,108 @@
...
+   */
+  public static async createPdf(payoutRequestId: number): Promise<PayoutRequestPdf> {
+    const payoutRequest = await PayoutRequest.findOne({
+      where: { id: payoutRequestId },
+      relations: ['requestedBy', 'approvedBy', 'payoutRequestStatus'],
+    });
+    if (!payoutRequest) return undefined;
+
+    const params = this.getPdfParams(payoutRequest);
+    return this.pdfGenerator.client.generatePayout(params).then(async (res: FileResponse) => {
```
Yoronex, line 100: Please use a try catch block here instead. In https://github.com/GEWIS/sudosos-backend/pull/205/commits/a477c0801f134c253c941ee0cf1ca27ac00e0ff1, I also changed it, because due to incorrect parameters the `.generatePayout()` function did not return a promise but an empty object. This results in an error that you cannot call .then on undefined, which is a really, really weird error. Awaiting the generation and putting everything in a try-catch block will result in better errors.

Yoronex (changes requested):

Yoronex (approved):

PR #198, opened 2024-07-03
ci: attempt to fix coverage base reporting

JustSamuel:
# Description
As discussed in the group chat, this should possibly fix the base reporting of the coverage.

## Related issues/external references
#32

## Types of changes
- CI/CD _(Changes to the CI/CD configuration)_

PR #209, opened 2024-07-11
ci: warn instead of fail if base coverage is not found.

JustSamuel:
# Description
image
## Types of changes


Yoronex (approved):

tomudding:
> One question though; how will it work if no artifacts exist (e.g. we deleted them from all previous workflow runs)? https://github.com/GEWIS/sudosos-backend/pull/198#pullrequestreview-2156476030

PR #210, opened 2024-07-12
fix: payout request pdf was always regenerated

JustSamuel:
# Description
Missing `eager` caused the pdf to always be null and thus always be regenerated.

## Types of changes


JustSamuel:
Note @Yoronex that this should be in main before we deploy because otherwise it pollute everything with payout requests lol

Maybe make an option to clear the directory and dump the table? you can always regenerate

Yoronex (approved):

PR #211, opened 2024-07-13
feat: add report pdf service

JustSamuel:
# Description
Adds the ability to turn fine reports into PDFs. Reports, by design, are not saved and directly returned.

## Types of changes


src/service/report-pdf-service.ts
```diff
@@ -0,0 +1,115 @@
...
+          vatAmount: PDF_VAT_HIGH,
+          // is actually unused
+          vatCategory: VAT.ZERO,
+          quantity: report.waivedCount,
+        }),
+      });
+      fines.push(waived);
+    }
+    const inclVat = report.handedOut.getAmount() - report.waivedAmount.getAmount();
+    const exclVat =  Math.round(inclVat  / (1 + (PDF_VAT_HIGH / 100)));
```
Yoronex, line 78: Heeft de Kasco al een conclusie wat we willen doen met de btw over boetes en hoe we dat willen rapporteren?

src/service/report-pdf-service.ts
```diff
@@ -0,0 +1,115 @@
...
+  TotalPricing,
+  VAT,
+} from 'pdf-generator-client';
+import { FineReport } from '../controller/response/debtor-response';
+import { PDF_GEN_URL, PDF_VAT_HIGH, UNUSED_NUMBER, UNUSED_PARAM } from '../helpers/pdf';
+
+const HANDED_OUT_FINES = 'Handed out';
+const WAIVED_FINES = 'Waived';
+
+export default class ReportPdfService {
```
Yoronex, line 37: Is this class only for fine reports? Because then the name should be different

JustSamuel: My intention is to have it for all reports in pdf

src/service/report-pdf-service.ts
JustSamuel: Ja, gewoon 21% en inboeken.

Yoronex (approved):

PR #212, opened 2024-07-13
ci: removed qemu setup

JustSamuel:
# Description
Since we have been rate limited we will just remove QEMU.

## Types of changes

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -26,9 +26,6 @@ jobs:
           tags: |
             type=ref,event=pr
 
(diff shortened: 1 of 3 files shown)


rinkp (approved):
Not requesting changes, but the rate limit does not just affect qemu. While removing some packages may help, it is not guaranteed. Additionally, if the GitHub run now takes less time, more jobs by others may run on the same IP resulting in the same rate limit notices.

This change should not influent the build process with the current build parameters (i.e. amd64), but will break arm64 compatibility in the buildx process if that is enabled (such as for osticket-dockerized)

Issue #170, opened 2024-04-24
Better permissions


SuperVK:
A few questions that need answers:
1. Should all BAC be able to edit transactions? (or only view them?)
2. Should Board be able to admin all users? (or only view them?)
3. Should all BAC be able to see all users? (and also edit them?)
4. Feel free to add more.

Furthermore -> Frontend should implement the permissions that the backend lays out, to prevent security through obsecurity, and make it clear what the current permissions are.

Goal of this issue: Answers these questions and implement them in the backend.

SuperVK:
Okay from my part:
1. BAC can see (and edit) transactions made by users, but cannot see the users themselves, what is the use of this?
2. I think should Board should have permissions to see all users, in case the BAC PM is not available to check certain things, I think it is nice to have the board as feedback.
3. I don't see the use of the whole BAC seeing all of the users, like is the case right now.
4. I cannot make a payoutrequest as a regular user? Should that not be changed?

SuperVK:
Concensus seems towards: BAC role can be removed, Seller role will receive more permissions, such as the creation of product, and the assigning of products to containers. Creation of containers and POS' will stay at the BAC PM.

Issue #166, opened 2024-04-17
Production RBAC roles and permissions are not tested


JustSamuel:
In the test suite we define roles with permissions per file, which means the actual roles and permission used in production are not tested.

Not tested here means that we never check that a LOCAL_USER should have the rights in production to update their own mail, or preferences. 

To fix this we should load the production RBAC in the test suite and rewrite some cases.

PR #215, opened 2024-07-22
feat: rbac in database

Yoronex:
# Description

This PR moves the role definitions with their permissions from the source code to the database. Roles can now be created/changed more easily, as changes to the codebase are no longer necessary to change GEWIS-specific roles.

SudoSOS now has two types of roles, namely system default roles and instance specific roles.
- System default roles are defined in the source code (in `/src/rbac/default-roles.ts`. These roles are `UserType`-specific and contain the most basic functionality that is required for that role to use SudoSOS. For example, everyone should be able to fetch their balance, local users should be able to change their local password, invoice users should be able to fetch all their invoices, and local admins should be able to do anything. These roles and their permissions are immutable during runtime. They can only be changed in the source code. Even if the role is changed in the database, SudoSOS will synchronize these roles in the database with their source code counterparts.
- Instance specific roles are the roles we all know and love. These are GEWIS specific roles like BAC, BAC PM, Board, Audit, etc. These roles are no longer defined in the database, but part of this PR's migration. When the roles have to change, the Superadmin can do this (or someone else with the correct permissions). When new permissions are introduced, the permissions for these instance specific can also be changed by using migrations (but do note that the role might not necessarily have to exist).

Moving the role definitions to the database brings multiple benefits for future functionality.
- The system default roles can be used in test suites to test whether a certain user indeed has the correct permissions. Some permission testing is now possible, but not for these instance specific roles. Right now, this methodology is only used in the `rbac-controller`, but it serves as a nice example. (https://github.com/GEWIS/sudosos-backend/issues/166)
- Roles can now more easily serve multiple purposes, as they can be created much more easily. Think of the event schedules that require people to be a member of a role. Or link a Role to a POS, allowing all members of that role to create transactions in that POS for anyone (scenario: BAC feuten now create transactions in name of a BAC member during borrels, but then they can also create transactions in name of themselves).
- The backend now returns a list of all permissions next to a list of roles, so the frontend can show functionality based on the rights the user has. (https://github.com/GEWIS/sudosos-backend/issues/170)

There are however a few "breaking" changes, as the responses of some endpoints have changed. However, they do not seem to be used by the frontend at the time of writing.
- `RoleResponse` now contains all the new attributes of a role. Furthermore, the attribute `role` has been renamed to `name`.
- The `roles` attribute in `AuthenticationResponse` is now deprecated in favor of the new `rolesWithPermissions`, as the latter also contains the permissions each role brings. Please use this new attribute @SuperVK.

Note that now, the access policy is checked against the database, and no longer a list of roles stored in the `RoleManager`. In theory, this will make each API request slightly slower, but given the performance of MariaDB, I do not deem this an issue. An alternative would be to cache all the permissions, but this will make the implementation quite complex. Also, in production SudoSOS is using multiple workers, meaning that changes in the permissions will take some time to take effect in each worker. If the database is broken somehow and checking the policy fails, the backend will return an HTTP 500 (see added test case in `policy-middleware.ts`).

Small point of discussion: now, we validate permissions based on the roles that are given. However, we can now also use the user ID to fetch the roles and permissions of the user in real time. We can then remove the roles from the JWT token, which makes new role assignments take effect immediately. If we want to gradually move to this alternative check, we have to create a new `userCan` method in the `RoleManager`, in which we give the user object instead of a list of roles. We then deprecate the old `can` method.

No explicit issue for this PR, but it is groundwork for many others.

## Types of changes

- Breaking change??? _(fix or feature that would cause existing functionality to change)_


Yoronex:
I seem to have forgotten some stuff with the migrations, but I'll look at that later today or tomorrow...

src/service/authentication-service.ts
```diff
@@ -137,18 +127,19 @@ export default class AuthenticationService {
    * @param token - The JWT token that can be used to authenticate.
    * @returns The authentication response.
    */
-  public static asAuthenticationResponse(
+  public static async asAuthenticationResponse(
```
JustSamuel, line 130: Why is this async?

src/service/event-service.ts
```diff
@@ -252,7 +253,7 @@ export default class EventService {
   public static async getSingleEvent(id: number): Promise<EventResponse | undefined> {
     const event = await Event.findOne({
       where: { id },
-      relations: ['createdBy', 'shifts', 'answers', 'answers.user'],
+      relations: { createdBy: true, shifts: { roles: true }, answers: { user: true } },
```
JustSamuel, line 256: maybe kinda out of scope of this PR? I'll allow it though.

src/service/rbac-service.ts
```diff
@@ -17,42 +17,246 @@
...
       };
     });
   }
+
+  /**
+   * Converts the RoleDefinitions object to an Roleresponse, which can be
+   * returned in the API response.
+   * @param roles - The role definitions to parse
+   */
+  public static asRoleResponse(roles: Role[]): RoleResponse[] {
```
JustSamuel, line 91: Maybe make it of type `Role => RoleResponse` and use a .map instead?

src/service/rbac-service.ts
```diff
@@ -17,42 +17,246 @@
...
+          return entDef;
+
+        }, {});
+      return permDef;
+    }, {});
+  }
+
+  /**
+   * Get a tuple with a list of all roles and the total number of
+   * roles matching the parameters
```
JustSamuel, line 154: Why the count?

src/service/rbac-service.ts
```diff
@@ -17,42 +17,246 @@
...
+   */
+  public static async updateRole(roleId: number, params: UpdateRoleParams): Promise<Role> {
+    const role = await Role.findOne({ where: { id: roleId } });
+    if (role == null) throw new Error('Role not found.');
+    if (role.systemDefault) throw new Error('Cannot update system default role.');
+
+    role.name = params.name;
+    await role.save();
+
+    const [[r]] = await this.getRoles({ roleId, returnPermissions: true });
```
JustSamuel, line 185: still kinky

src/service/rbac-service.ts
```diff
@@ -17,42 +17,246 @@
...
+   * @param permissionRule
+   */
+  public static async removePermission(roleId: number, permissionRule: Omit<PermissionRule, 'attributes'>) {
+    const matches = (await Permission.find({ where: {
+      roleId,
+      entity: permissionRule.entity,
+      action: permissionRule.action,
+      relation: permissionRule.relation,
+    }, relations: { role: true } }));
+    if (matches.length === 0) {
```
JustSamuel, line 229: maybe then assign it to a const called `match`, I find array indexing, even if its with 0, to be ugly and prefer to swap it out as early as possible. I.e prefer a const instead of `matches[0]`

JustSamuel (changes requested):
Not all are required, just some comments.

src/controller/rbac-controller.ts
```diff
@@ -42,28 +47,60 @@ export default class RbacController extends BaseController {
...
+          handler: this.updateRole.bind(this),
+          body: { modelName: 'UpdateRoleParams' },
+        },
+        DELETE: {
+          policy: async (req) => this.roleManager.can(req.token.roles, 'delete', 'all', 'Role', ['*']),
+          handler: this.deleteRole.bind(this),
+        },
+      },
+      '/roles/:id(\\d+)/permissions': {
+        POST: {
```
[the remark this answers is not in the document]
Yoronex: You can add permissions in bulk (as the POST request takes an array of permission rules), but you can only delete them one at a time. This is because - with a POST request - it is easy to allow an array of objects as a request body, but DELETE requests do not have such a body, so you need to put everything in the URL.

src/controller/rbac-controller.ts
```diff
@@ -73,4 +110,263 @@ export default class RbacController extends BaseController {
...
+      const roleId = Number(id);
+      const request = { ...body } as UpdateRoleParams;
+
+      const validation = await verifyUpdateRoleRequest(request);
+      if (isFail(validation)) {
+        res.status(400).json(validation.fail.value);
+        return;
+      }
+
+      let [[role]] = await RBACService.getRoles({ roleId }, { take: 1 });
```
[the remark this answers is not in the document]
Yoronex: This indeed looks a bit weird. However, I was trying something new regarding the way we handle pagination in the service methods. I do not remember in which issue/PR we discussed this, but somehow we want to incorporate pagination in a service method, but not have this method return actual responses. Therefore, the `getRoles` method returns the same object type as the `findAndCount` method of TypeORM, namely a tuple of records and the total number of records that match the search criteria (i.e. the count). If you want only the first found record, you indeed do this to get the first element (don't immediately remember the name of this shorthand method).

src/rbac/default-roles.ts
```diff
@@ -0,0 +1,277 @@
...
+      },
+    }];
+  }
+
+  /**
+   * Synchronize a single default role with the database
+   * @param roleDefinition
+   * @private
+   */
+  private static async synchronizeRole(roleDefinition: DefaultRole): Promise<Role> {
```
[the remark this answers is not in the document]
Yoronex: I changed some logic in the if statement and added comments to explain what I'm doing, so hopefully this makes it more readable.

src/service/authentication-service.ts
Yoronex: Leftover from an earlier (uglier) implementation, fixed now.

src/service/event-service.ts
Yoronex: I did add an extra relation to be returned (`shifts.roles`), so I thought "why then not also immediately fix the ugly array syntax" so here we are.

src/service/rbac-service.ts
Yoronex: See the comment above about pagination for this method.

src/service/rbac-service.ts
Yoronex: See the comment above about pagination for this method.

src/service/rbac-service.ts
Yoronex: Changed it!

src/controller/rbac-controller.ts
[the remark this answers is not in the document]
JustSamuel: That should work, but I still think that the `count` is something you don't always need, so seems just weird that it is always returned.

JustSamuel (approved):

PR #216, opened 2024-07-23
fix: debtor test case inconsistent test results

Yoronex:
# Description

The test case `should correctly put two fines in same userFineGroup` sometimes fails, because the internal state has changed in the previous test case. The additions are now properly cleaned up.

## Types of changes


JustSamuel (approved):
nice

PR #218, opened 2024-07-24
fix: migration table missing when seeding database

Yoronex:
# Description

Migration table is missing when running `npm run seed`, which makes it difficult to test migrations.

## Types of changes


JustSamuel (approved):
seems fine

PR #222, opened 2024-07-25
feat: POS cashier roles

Yoronex:
# Description

Adds a many-to-many relation between POSs and roles, allowing to associate groups of users with POSs. There already exist a way to do this, namely via ownership using the `MemberAuthenticator`. However, this is ownership, so all users are (in theory) able to make changes to the POS. 

This PR allows a second group of users to be linked to a POS, namely "cashiers". In practice, this will be the BAC feuten, so they can also create transactions in their own name during borrels. Right now, they do this in the name of a BAC member, which is confusing. To keep changes to SudoPOS minimal, there is a new endpoint that returns all users associated with a POS, seperated by their relation with the POS.

## Types of changes

### DIFF
--- a/src/controller/point-of-sale-controller.ts
+++ b/src/controller/point-of-sale-controller.ts
@@ -39,6 +39,7 @@ import {
 import userTokenInOrgan from '../helpers/token-helper';
 import TransactionService from '../service/transaction-service';
 import { PointOfSaleWithContainersResponse } from './response/point-of-sale-response';
+import UserService from '../service/user-service';
 
 export default class PointOfSaleController extends BaseController {
   private logger: Logger = log4js.getLogger('PointOfSaleController');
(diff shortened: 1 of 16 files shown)


Yoronex:
Implements https://github.com/GEWIS/sudosos-backend/issues/223

src/controller/point-of-sale-controller.ts
```diff
@@ -361,6 +368,49 @@ export default class PointOfSaleController extends BaseController {
...
+        where: { id: pointOfSaleId },
+        relations: { cashierRoles: true, owner: true },
+      });
+      if (!pos) {
+        res.status(404).json('Point of Sale not found.');
+        return;
+      }
+
+      const { owner } = pos;
+      const ownerMembers = await UserService.getUsers({ organId: pos.owner.id });
```
JustSamuel, line 399: I still think that a 'User.find(options)` would be better, since you can ommit the whole .records, but I don't think this has to change.

JustSamuel (approved):

PR #224, opened 2024-07-25
feat: add vat to transfers

JustSamuel:
# Description
Since `WriteOffs` and `Inactivity Fees` have a VAT percentage but are not products, we have a need for adding VAT groups to transfers. This PR implements the bases of this. 

## Types of changes

### DIFF
--- a/src/controller/request/transfer-request.ts
+++ b/src/controller/request/transfer-request.ts
@@ -23,9 +23,8 @@ import { DineroObjectRequest } from './dinero-request';
  * @typedef {object} TransferRequest
  * @property {string} description.required - Description of the transfer.
  * @property {DineroObjectRequest} amount.required - Amount of money being transferred.
- * @property {integer} type.required - Type of transfer.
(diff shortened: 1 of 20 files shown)


SuperVK:
We cannot merge it like this, the breaking change will be a nightmare to deploy correctly at exactly the same time (and yes this is a breaking change as it breaks the frontend if we change nothing)

SuperVK:
If we follow semantic versioning, this should be a v2, then we could switch to `sudosos.gewis.nl/api/v2`, and then the frontend can update on its own

Yoronex (approved):

Yoronex:
You can also temporarily include both `amount` and `amountInclVat` in the response, as this gives you a window to migrate the frontend to use the new attribute. I have done the same with the new roles and permissions.

JustSamuel:
@SuperVK @Yoronex I added the `amount` property back, but deprecated. To be removed Soon TM.

Issue #106, opened 2023-11-01
Change existing user's usertype


Yoronex:
Some _very_ old GEWIS members still wanted to use SudoSOS last Friday, but they couldn't, as their GEWIS account has been terminated and deleted from the database. Their accounts can no longer be recovered, so logging in via GEWIS is no longer possible. However, these accounts do still exist within SudoSOS.

To fix this use case, we should implement some functionality to change the user type of accounts. More specifically, we should be able to change a GEWIS user to a local user. The user should receive an email about this change with a link to change their password. 

For completeness, we should also implement a function to do this the other way around: in case someone with a SudoSOS account gets accepted into GEWIS by the board as an external member. We should then be able to convert the local account to a GEWIS account.

For now, we can do both cases via some direct database manipulations, but it is much nicer to implement functionality for this.

Yoronex:
There is some controversy regarding the responsibility for reactivating old GEWIS/SudoSOS accounts, so this will be discussed at the coming ABC meeting.

Yoronex:
Decided at ABC meeting 21 (17/11/2023) that we are going to implement this.

PR #229, opened 2024-07-26
[Draft] feat: add updating a user's type

Yoronex:
# Description

Add `type` to the `UpdateUserRequest`. @SuperVK does the frontend use integers to denote user types, or strings?

Implements https://github.com/GEWIS/sudosos-backend/issues/106

## Types of changes

### DIFF
--- a/src/controller/user-controller.ts
+++ b/src/controller/user-controller.ts
@@ -746,29 +746,19 @@ export default class UserController extends BaseController {
    * @return {string} 400 - Bad request
    */
   public async updateUser(req: RequestWithToken, res: Response): Promise<void> {
-    const body = req.body as UpdateUserRequest;
     const parameters = req.params;
-    this.logger.trace('Update user', parameters.id, 'with', body, 'by user', req.token.user);
-
-    if (body.firstName !== undefined && body.firstName.length === 0) {
-      res.status(400).json('firstName cannot be empty');
-      return;
-    }
-    if (body.firstName !== undefined && body.firstName.length > 64) {
-      res.status(400).json('firstName too long');
-      return;
-    }
-    if (body.lastName !== undefined && body.lastName.length > 64) {
-      res.status(400).json('lastName too long');
-      return;
-    }
-    if (body.nickname !== undefined && body.nickname.length > 64) {
-      res.status(400).json('nickname too long');
-      return;
-    }
-    if (body.nickname === '') body.nickname = null;
+    this.logger.trace('Update user', parameters.id, 'with', req.body, 'by user', req.token.user);
 
     try {
+      const validation = await verifyCreateUserRequest(req.body);
+      if (isFail(validation)) {
+        res.status(400).json(validation.fail.value);
+        return;
+      }
+
+      const body = req.body as UpdateUserRequest;
+      if (body.nickname === '') body.nickname = null;
+
       const id = parseInt(parameters.id, 10);
       // Get the user object if it exists
       let user = await User.findOne({ where: { id, deleted: false } });
@@ -778,13 +768,8 @@ export default class UserController extends BaseController {
         return;
       }
 
-      user = {
-        ...body,
-      } as User;
-      await User.update(parameters.id, user);
-      res.status(200).json(
-        await UserService.getSingleUser(asNumber(parameters.id)),
-      );
+      const response = await UserService.updateUser(id, req.body);
+      res.json(response);
     } catch (error) {
       this.logger.error('Could not update user:', error);
       res.status(500).json('Internal server error.');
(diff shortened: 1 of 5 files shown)


SuperVK:
Frontend uses the strings it gets from the backend, example:

https://github.com/GEWIS/sudosos-frontend/blob/develop/apps/dashboard/src/modules/user/components/ChangePassword.vue#L43

src/controller/user-controller.ts
```diff
@@ -778,13 +768,8 @@ export default class UserController extends BaseController {
...
       }
 
-      user = {
-        ...body,
-      } as User;
-      await User.update(parameters.id, user);
-      res.status(200).json(
-        await UserService.getSingleUser(asNumber(parameters.id)),
-      );
+      const response = await UserService.updateUser(id, req.body);
```
JustSamuel, line 771: This won't work that easily. 

Going from a GEWIS Member to a Local User should also remove the GEWIS member info right? The LDAP integration is also based on GEWIS m-numbers.

In general there is quite some tricky stuff, since there are 2 ways accounts are created in SudoSOS: 
- A new user is created by logging in using the GEWIS website (and the account is bound to the m-number)
- A new user is created when logging in using LDAP and no account associated with the m-number exists.

This means that removing someones MEMBER type status MUST only be done when you are sure they don't have a GEWIS account anymore. 

A solution could be to keep the GEWIS m-number info stored, but just throw an error if someones tries to log in based on m-number whilst its not a type MEMBER? (I _believe_ this is almost the current behavior). 

@tomudding would it be bad if we kept the m-number to SudoSOS id info stored?

JustSamuel (approved):

src/controller/user-controller.ts
tomudding: > @tomudding would it be bad if we kept the m-number to SudoSOS id info stored?

Ideally, we never keep it to prevent people from re-associating the transaction data to the user once we start deleting data (i.e. the yet to be implemented anonymisation process - though note that people can technically already request this using the GDPR).

However, at this point, we are not yet deleting the data but only changing types, as such, I think it is fine to keep track of the user's m-number if it exists. Preventing token creation/login when trying to use a MEMBER style login is also logical, but you will have to make sure that this is properly documented within the codebase (and properly handle it in both the dashboard and POS)!

Yoronex: So what do we want to do? Because the workflow seems much more complex than I imagined. I personally don't think we should make a dependency on the GEWIS database to verify whether MEMBER accounts are actually deleted from the GEWIS database, because it seems unnecessary? I also think that keeping m-account related information might not cause any harm, because that would make changing the UserType a revertible change, and a not a destructive one.

JustSamuel: Keep the data, for now, but maybe add the following:

The main use case for switching UserType is from MEMBER to LOCAL, however a LOCAL user needs a password. So I say we send the person an info mail with a (re)set password link? And maybe add some integration test cases that a new account wont be created in both the cases when the m-number still exists but instead errors?

This might be a bit more involved, so let me know if you still want to do this.
When I looked at it, just adding an extra check to the endpoints in the `gewis-authentication-controller` [here](https://github.com/GEWIS/sudosos-backend/blob/e8cf1d272e989ca62463f2f48211c7ba5758c35c/src/gewis/controller/gewis-authentication-controller.ts#L118) might be a solid move, since that would throw a nice error. (return a 403, error m-number unregistered).

As a possible extra you could also throw an error [here](https://github.com/GEWIS/sudosos-backend/blob/e8cf1d272e989ca62463f2f48211c7ba5758c35c/src/gewis/gewis.ts#L43) since that is the service level of where the accounts get created.

Yoronex: I think I'll park this PR for now, in order to first also work on the improved SudoSOS mailings. If you want to @JustSamuel, you can also take a look at those other things that you mentioned, as you are more familiar with the GEWISDB and LDAP integrations (and these things are difficult to test)

rinkp: > I personally don't think we should make a dependency on the GEWIS database to verify whether MEMBER accounts are actually deleted from the GEWIS database, because it seems unnecessary?

@Yoronex It is necessary to avoid one person with two accounts. In case the BAC Treasurer converts a member to a local account and later the same member wants to join a GEWIS activity or view photos on the website so now the member is activated in GEWISDB either way. They then sign in to sudosos with that account and suddenly they have two accounts.
Remember that maintaining local accounts is very undesirable for people who reslly should just have become graduates, because now suddenly the bacpm has to make sure the personal data is always updated instead of the secretary of GEWIS

Yoronex:
This issue will not be addressed this summer. Let's put it back on the backlog for now.

Issue #226, opened 2024-07-26
[Feature] Server Settings


JustSamuel:
### What would you like?

Some settings or entities, such as VAT, only really exist on runtime. However, for fines, write-offs and other transfers we would like to attach the high VAT group to it. 

One solution would be to add which VAT group should be added to the endpoint, however this would allow for mistakes. (I.e, the BACPM must select the same VAT group every week when handing out fines.). A proposed feature is that we have `ServerSettings` which define non-secret but saved settings, such as: "Which VAT group Id is high VAT?".

Other stuff could be the LDAP settings, which are now in ENV but could also simply be stored in the database.

### How could it be implemented?

Similar to how `narrowcasting-core` has implemented it [here](https://github.com/GEWIS/narrowcasting-core/blob/577086a1f1ab4077d3efc59bff9a6d3075d640c1/src/modules/root/server-settings-store.ts#L11).

PR #231, opened 2024-07-27
feat: add server settings to store global settings in database

Yoronex:
# Description

Helper class to store system global settings in the database (instead of an environment variable). The store is synchronous in fetching settings, but this requires settings to be cached locally. Even though you can dynamically change these settings by using the `setSetting` method, the setting will then only change in that particular worker until SudoSOS is rebooted.

Or we can make the `getSetting` also asynchronous, which would solve this issue. What do you prefer?

Implements https://github.com/GEWIS/sudosos-backend/issues/226;

## Types of changes


JustSamuel (approved):
I think its fine to leave it sync

PR #232, opened 2024-07-27
feat: Point of Sale users to perform POS-related actions

Yoronex:
# Description

This PR adds a new special user type, namely "Point of Sale" this type of user will (and should) never be returned by any endpoints, but only exist internally. With this type of user, the POS frontend can perform privacy-sensitive actions (like fetching all users) with its own, special token, instead of the user's token. Then, we can also better secure endpoints like PIN authentication, but these will all be implemented in a later stage.

POS tokens behave just like any JWT token, except their expiry is much, much longer (2 weeks instead of 1 hour). This is necessary to prevent logouts over longer periods of time, like the Christmas break.

When this PR is merged, the next step is to implement the new POS authentication flow in the POS frontend:
- User logs in on the POS just like on the Dashboard: via GEWIS.nl or username/password.
- POS fetches all Points of Sale the logged in user can open. This is the same list as the POSs the user owns (either directly or via organ).
- User selects POS; an API call is made to the new endpoint with the user's token to create a POS token.
- POS token should be used to fetch all relevant data; User token should only be used to create a transaction.

When the POS has implemented this new flow, we can start with improving the lesser tokens:
- Lesser tokens should include the ID of the POS they have been authenticated at. This ensures the user can only create transactions on the POS it is authenticated at. Furthermore, we then also know whether the user is allowed to switch the POS, because we know whether they own the current POS as well.
- PIN, NFC and EAN authentications should also require JWT tokens, namely the token of the POS.
- Revoke the RBAC rights for regular users to fetch all other users, POSs, containers, products, and product categories.

First step for implementing https://github.com/GEWIS/sudosos-backend/issues/62

## Types of changes


src/authentication/jwt-expiry.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+export const DEFAULT_JWT_EXPIRY = 3600; // seconds (so 1 hour)
+
+// Two weeks (to also accommodate for things like holidays for public points of sale)
+export const POINT_OF_SALE_JWT_EXPIRY = 60 * 60 * 24 * 2;
```
JustSamuel, line 21: Maybe put this in the settings stores?

src/entity/point-of-sale/point-of-sale.ts
```diff
@@ -44,6 +44,10 @@ export default class PointOfSale extends BaseEntity {
   @ManyToOne(() => User, { nullable: false, eager: true })
   public owner: User;
 
+  @OneToOne(() => User, { nullable: false })
```
JustSamuel, line 47: OneToOne? So we are sure that every POS must have a unique user?

JustSamuel (approved):

src/entity/point-of-sale/point-of-sale.ts
Yoronex: I think it is good practice to do it that way, so you can always trace back who (or in this case: what) performed a certain action. If you use the same user everywhere, you cannot distinguish devices.

Issue #68, opened 2023-09-18
Change database module to create and provide a TypeORM DataSource object


Yoronex:
The current TypeORM database implementation still uses deprecated connections. It is possible to migrate over time to the new `DataSource` implementation as these are backwards compatible with `Connection`, but `database.ts` should first expose a DataSource object. This can be done either via a module or a Singleton class. The former would be preferred, as a database connection would then be opened once the module is loaded by some other module. Opening a database connection is however asynchronous, so whether this is actually possible should be looked into.

PR #233, opened 2024-07-27
chore: use TypeORM DataSource to initialize database

Yoronex:
# Description

Still initialize the database, but also expose a `DataSource` object that should be used over `getConnection()`.

Fixes https://github.com/GEWIS/sudosos-backend/issues/68.

## Types of changes


JustSamuel (changes requested):
TypeORM doesn't like this since you do not actually export an `instantiated` DataSource object.

Trying to use the CLI gives the following:

`PS C:\Users\Samuel\WebstormProjects\GEWIS\SudoSOS\sudosos-backend> typeorm migration:show -d .\out\src\database\database.js
Error during migration show:
Error: Given data source file must contain export of a DataSource instance
    at CommandUtils.loadDataSource (C:\Users\Samuel\WebstormProjects\GEWIS\SudoSOS\sudosos-backend\node_modules\typeorm\commands\CommandUtils.js:44:19)
    at async Object.handler (C:\Users\Samuel\WebstormProjects\GEWIS\SudoSOS\sudosos-backend\node_modules\typeorm\commands\MigrationShowCommand.js:27:26)
`
That is why I had the `export const AppDataSource = new DataSource(options);`

Yoronex:
But that export that you made was not usable in the actual codebase... When you imported AppDataSource and tried to use it (for example with `getRepository(<entity>)`, the DataSource would never be initialized.

JustSamuel:
> But that export that you made was not usable in the actual codebase... When you imported AppDataSource and tried to use it (for example with `getRepository(<entity>)`, the DataSource would never be initialized.

Yeah, but I believe that is how the CLI needs it ? As I see [here](https://orkhan.gitbook.io/typeorm/docs/data-source) you must always init the AppDataSource

Yoronex:
@JustSamuel can you verify the CLI works now?

JustSamuel (approved):
Works!

Issue #77, opened 2023-09-28
Fix Invoice Filtering Bug in `GET /invoices` Endpoint 


CodeNamedRobin:
**Description:**
Our API endpoint GET /invoices offers functionality to filter invoices based on their state, allowing users to retrieve invoices that fall into categories such as CREATED, SENT, PAID, or DELETED. However, there is a bug in this feature - regardless of the state provided in the request, the endpoint returns all invoices without proper filtering.

**Task:**
- Investigate the GET /invoices endpoint to identify the root cause of the filtering issue.
- Implement a fix to ensure that the endpoint correctly filters invoices based on the specified state parameter, returning only invoices that match the selected state.
- Update relevant documentation, if needed,  to reflect the corrected behavior of the endpoint.

**Expected Outcome:**
By addressing this issue, we aim to enhance the functionality and accuracy of our API by resolving the bug in the invoice state filtering. This will provide a more reliable and intuitive experience for our users, ensuring that they can retrieve the specific invoices they need.

**Skills Required:**
- Backend development
- Knowledge of API endpoints and query parameter handling
- Debugging and issue resolution

**Difficulty Level:**
Beginner

**Additional Information:**
This issue is an excellent opportunity for Hacktoberfest participants to contribute to our project by fixing a simple bug that impacts the invoice filtering feature. If you have any questions or need guidance on how to approach this task, please don't hesitate to ask. Your efforts will help improve the overall functionality and usability of our API.

JustSamuel:
Actually not as straight forward as I hoped.

`currentState` is a computed property, i.e. the state of the last InvoiceState. This means that filtering isn't very straight forward. 

A solution would most likely require rewriting the getInvoices into a queryBuilder which also isn't preferred.

PR #234, opened 2024-07-27
feat: invoice entity rework

JustSamuel:
# Description
Preface: I accidentally grew the scope of this PR.

This PR adds 2 extra columns to the `Invoice` entity. Namely `date` for, well, date, and `attention` for the extra attention row used when addressing companies. 

This PR also adds a `latestStatus` which is the latest status update of the Invoice. This is needed to do effective database queries on Invoice state, which I found to be a must when creating the frontend. I kept the endpoints the same, such that it integrates nicely.

TODO:
- [x] Migrations

#77 

## Types of changes

### DIFF
--- a/src/subscriber/invoice-status-subscriber.ts
+++ b/src/subscriber/invoice-status-subscriber.ts
@@ -27,20 +27,19 @@ export default class InvoiceStatusSubscriber implements EntitySubscriberInterfac
   }
 
   async afterInsert(event: InsertEvent<InvoiceStatus>): Promise<void> {
-    await InvoiceStatusSubscriber.updateInvoiceStatus(event.manager, event.entity.invoice.id, event.entity);
+    await InvoiceStatusSubscriber.updateInvoiceStatus(event.manager, event.entity.invoice.id);
   }
 
   async afterUpdate(event: UpdateEvent<InvoiceStatus>): Promise<void> {
-    await InvoiceStatusSubscriber.updateInvoiceStatus(event.manager, event.databaseEntity.invoice.id, event.databaseEntity);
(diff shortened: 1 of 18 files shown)


src/subscriber/invoice-status-subscriber.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  static async updateInvoiceStatus(manager: EntityManager, invoiceId: number, status: InvoiceStatus): Promise<void> {
+    const invoiceRepository = manager.getRepository(Invoice);
+    const invoice = await invoiceRepository.findOne({ where: { id: invoiceId }, relations: ['latestStatus'] });
+
+    if (invoice) {
+      invoice.latestStatus = status;
+      await invoiceRepository.save(invoice);
+    }
+  }
```
Yoronex, line 45: Does this function always return the latest status? What if you update a status? Because then that one will become the latest? What if this status is deleted? Is that case also covered?

Yoronex (changes requested):

src/subscriber/invoice-status-subscriber.ts
JustSamuel: There are test cases for updating a status (multiple times) and those pass.
You cannot update nor delete a status, we don't have options for those.

But if you would be to update a status (why?) then that would become the latest status. This _might_ not be the cleanest solution, but it is way better than not being able to effectively filter on them.

Yoronex: Follow up question: would it then not make sense to include such an extra check? Might indeed take an extra query and a loop to see whether it is indeed the most current status, but it would make the implementation more robust. Let's say we want to implement status "comments" in the future (like PP has), then those changes might cause havoc when you make a typo and fix that.

Yoronex (approved):

Issue #156, opened 2024-02-15
Refactor mailing tempalte to be in the GEWIS STIJL


JustSamuel:
Based on @rinkp his [remark on this issue](https://github.com/GEWIS/sudosos-backend/issues/144#issuecomment-1904813073). It might be nice to change the mailing theme.

Issue #136, opened 2024-01-12
Typo in the mail


SuperVK:
There is a slight typo in https://github.com/GEWIS/sudosos-backend/blob/develop/src/mailer/templates/user-debt-notification.ts#L60 where accoring should be according.

rinkp:
Please then also fix 

[into the future](https://github.com/GEWIS/sudosos-backend/blob/develop/src/mailer/templates/user-debt-notification.ts#L62C1-L62C122)
It would be pretty cool to time travel, but maybe not as a punishment

```diff
-<p>Go to the SudoSOS website to deposit money into your account. With this you prevent getting fined into the future.</p>
+<p>Go to the SudoSOS website to deposit money into your account. With this you prevent getting fined in the future.</p>
```

Issue #144, opened 2024-01-20
Sending receipts and balance notifications


rinkp:
Perhaps we may want to consider: sending receipts / notifications

Currently people discover their balance is negative when they receive a notification that they will be fined, not when they actually did a transaction. While this works for users that maintain a close-to-negative balance, this does not work for all users. 

My proposal is to:
- [ ] Allow users to enable transaction notifications including a summary of bought items for all transactions, i.e. like a receipt
- [ ] Allow users to enable balance notifications each first transaction of the day while their balance is <€0

CodeNamedRobin:
People do get notifications exactly when they go in debt, however this might still be a nice idea to implement.

JustSamuel:
Clarification:

Do you propose that the user receives a "transaction" overview as a receipt? 
Don't people already get balance notifications when dropping down to <0?

rinkp:
> Do you propose that the user receives a "transaction" overview as a receipt?

Yes, if they enable this. This allows users to immediately notice something has been charged to their account while they were not aware of it. I propose a 3-state selector.

Do you want to receive notifications for transactions:
- Never
- When someone else charges my account (including social drinks)
- For all transactions

> Don't people already get balance notifications when dropping down to <0?

@CodeNamedRobin told me they do, but the point is that you will only find out your account has been erroneously charged once you have a debt which I think is not nice.

rinkp:
@JustSamuel, the example you asked for

![example email](https://github.com/GEWIS/sudosos-backend/assets/17340320/36e2f561-6137-480e-8686-b0791c4f388b)

```yaml
# GEWIS Newsletter
---
# All mailings require some settings. These can be set here. Don't forget to set the title ("Algemene mededeling; tekst zelf invoeren")
version: 1.1
settings:
    title: "Your SudoSOS transaction on January 1st"
    email: "sudosos@removed-for-spam-reasons"
    #unsubscribe: https://lists.gewis.nl/postorius/lists/gewis.lists.gewis.nl/
    lang: en-UK
    #lang: nl-NL
    #url: "https://secr.personal.gewis.nl/newsletter/2022/newsletter-8.html"
    follow: false


# A mailing consists of a series of messageblocks, each with their own type
# The available types are "header", "topblock", "activities", "activityfooter", "extramessage" and "tinyfooter"
messageblocks:

# Each mailing must have a header to make the template shine! It contains the title and the date.
    - type: "header"
      title: "SudoSOS update"
[...]

PR #236, opened 2024-07-29
feat: GEWIS email template

Yoronex:
# Description

All HTML emails now use the GEWIS email template. It is still the red one, colour is yet to be decided before we merge to main. I also slightly refactored the way emails are built while I was at it.

Implements https://github.com/GEWIS/sudosos-backend/issues/156 and https://github.com/GEWIS/sudosos-backend/issues/136.

## Types of changes


Yoronex:
Before we deploy this PR in production, we need to have a decision on the colouring scheme for SudoSOS and/or its mailings. (@JustSamuel @SuperVK @CodeNamedRobin). For now, it would be nice if we can merge it, so I can also work on issues like https://github.com/GEWIS/sudosos-backend/issues/144.

JustSamuel (approved):

Yoronex:
@SuperVK was there perhaps already a conclusion from the poll you sent to the ABC?

SuperVK:
GEWIS Rood: 13 points
SudoSOS Goudgeel: 11 points
BAC Groen: 12 points

tbh too close to call in this type of way

image

PR #243, opened 2024-07-30
fix: mistake in updating invoice

JustSamuel:
# Description
Turns out I do monkey paste.

## Types of changes


src/service/invoice-service.ts
```diff
@@ -397,16 +397,26 @@ export default class InvoiceService {
...
+          postalCode,
+          city,
+          country,
+          reference,
+          attention,
+          date: date ? new Date(date) : undefined,
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        }).filter(([_, value]) => value !== undefined),
+      ),
+    );
```
Yoronex, line 417: Why not just do `await Invoice.update({ id: base.id }, update)`? You only still need to parse the date to an actual date, but that method looks much cleaner and easier to understand to me.

Yoronex (changes requested):

src/service/invoice-service.ts
JustSamuel: You would still need to filter undefinea

JustSamuel: Heb een nieuwe suggestie gedaan voor je

Yoronex: But is `UpdateInvoiceParams` not a "subset" of `Invoice`? Then you can just use `Invoice.update` method, as undefined objects should be ignored? Only if they are `null`, they should be set to `null` in the database. For any attributes that still need to be parsed (like date), you can use the spread operator, like

```
const { date, ...props } = update;
await Invoice.update({ id: base.id }, { ...props, date: date ? new Date(date) : undefined });
```

JustSamuel: that works, thanks!

Yoronex (approved):

PR #245, opened 2024-08-03
refactor(payoutrequests): move to `getOptions` method

JustSamuel:
# Description
Refactors the `payout-request-service` and fixes some bugs

@Yoronex in the invoice setup we use an extra column, here we use a subquery to filter on current states. What is the better setup? Since the invoice needed a migration I suggest not to change it.

## Types of changes

### DIFF
--- a/src/controller/payout-request-controller.ts
+++ b/src/controller/payout-request-controller.ts
@@ -239,6 +239,14 @@ export default class PayoutRequestController extends BaseController {
       return;
     }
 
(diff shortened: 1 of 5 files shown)


Yoronex (approved):

Yoronex:
I prefer to not cache the status of an entity, as this "cache" violates the normalized database. I know we do already have some violations here and there, but less is more in my opinion. I am also aware that such a greatest-n-per-group query makes querying more difficult, but this outweighs the caching thing imo.

Issue #197, opened 2024-07-03
[Feature] Automagic Invoices


JustSamuel:
### What would you like?

Create automatic invoices for enabled users. These invoices will be more like receipts and should make it easier for the BACPM.

### Why is this needed?

- Other clubs don't have to wait for the BAC treasurer to eventually decide to send an invoice, because SudoSOS will send the invoice immediately at night.
- The BAC treasurer doesn't have to make petty invoices of a few euros and invoices.
- The BAC treasurer no longer has to look after all these accounts.

### How could it be implemented?

Automatic invoices will be separate from normal invoices. The BACPM can still make manual invoices (for things such as CEB account).

The implementation will be that automatic invoices should be counted in the "transactionOverview"  where as manual invoices are not. This could be tricky, so this should be done carefully.

We can use an internal reference on the invoice (SUDOSOSXXXX) which will make it clear that it is an automatic invoice. It was also suggested that an automatic invoice also create a "paymentRequest" from stripe such that an account can instantly pay using stripe. This will allow us to send payment links (hopefully that doesn't kill our spam score)

But the main gist is to turn automatic invoices into automatic "receipts" which will be paid for through SudoSOS. This also means that they do not have to be incorporated into the BAC bookkeeping as they will be reported within the transactionOverview.

tomudding:
> It was also suggested that an automatic invoice also create a "paymentRequest" from stripe such that an account can instantly pay using stripe. This will allow us to send payment links (hopefully that doesn't kill our spam score)

What would the payment links look like? I assume you will get some URL that looks like `{sudosos_domain}/invoice/{long_random_string}` that prompts you with the interface from the `paymentRequest`?

How will you handle expiration of URL/invoice? How will you deal with accidental double payments because Stripe's webhooks are delayed?

JustSamuel:
> > It was also suggested that an automatic invoice also create a "paymentRequest" from stripe such that an account can instantly pay using stripe. This will allow us to send payment links (hopefully that doesn't kill our spam score)
> 
> What would the payment links look like? I assume you will get some URL that looks like `{sudosos_domain}/invoice/{long_random_string}` that prompts you with the interface from the `paymentRequest`?
> 
> How will you handle expiration of URL/invoice? How will you deal with accidental double payments because Stripe's webhooks are delayed?

[...]

PR #253, opened 2024-08-12
feat: separate invoice-transaction relation to support debit and credit invoices

Yoronex:
# Description

Changes the relation between invoices and subtransactionrows, such that every subtransactionrow can have both a debit invoice and a credit invoice. This was not possible before, but fortunately we have not yet encountered this issue in production (yet).

Step one of https://github.com/GEWIS/sudosos-backend/issues/197.

## Types of changes

### DIFF
--- a/src/controller/request/validators/invoice-request-spec.ts
+++ b/src/controller/request/validators/invoice-request-spec.ts
@@ -49,7 +49,10 @@ import Invoice from '../../../entity/invoices/invoice';
 async function validTransactionIds<T extends BaseInvoice>(p: T) {
   if (!p.transactionIDs) return toPass(p);
 
-  const transactions = await Transaction.find({ where: { id: In(p.transactionIDs) }, relations: ['from', 'subTransactions', 'subTransactions.subTransactionRows', 'subTransactions.subTransactionRows.invoice', 'subTransactions.to'] });
+  const transactions = await Transaction.find({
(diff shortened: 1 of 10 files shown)


JustSamuel:
It is very tricky for a subTransactionRow to be both connected to a credit and debit invoice. 

The reason for this is that an Invoice inherently means that the payment or payout of a certain transaction will be done outside of SudoSOS. 

A debit Invoice means that a User will pay for the Transactions outside of SudoSOS.
A credit Invoice is actually being used as Payout with linked transactions. As the creation of a credit invoice decreases the sellers balance by that amount.

This does make things complicated when you try and Invoice transactions which have been paid out in the form of a credit Invoice. 

Take the following example:
- I.V.V sells 10 euros of wine to an invoice account
- I.V.V pays out his balance by requesting a credit invoice of the wine
- The invoice account makes an invoice for that wine

The credit invoice is a form paying out the balance. The invoice is a form of saying "hey this transaction will be settled outside of SudoSOS". So I.V.V having a balance of negative -10 makes sense, they received that money twice.

Yoronex:
Closed because it is replaced by PRs https://github.com/GEWIS/sudosos-backend/pull/259 and https://github.com/GEWIS/sudosos-backend/pull/266 (see https://github.com/GEWIS/sudosos-backend/issues/260).

PR #254, opened 2024-08-12
feat: split StripePaymentIntent and StripeDeposit entities

Yoronex:
# Description

Creates a separate table of payment intents, so automatic invoices can also use this table and thus all Stripe transactions are handled the same. Depends on https://github.com/GEWIS/sudosos-backend/pull/253.

## Related issues/external references

Step two of https://github.com/GEWIS/sudosos-backend/issues/197.

## Types of changes

- New feature _(non-breaking change which adds functionality)_

PR #257, opened 2024-08-13
feat(stripe): separate Stripe payment intents from StripeDeposit entity

Yoronex:
# Description

Standalone of https://github.com/GEWIS/sudosos-backend/pull/254.

Prework for https://github.com/GEWIS/sudosos-backend/issues/197.

## Types of changes

### DIFF
--- a/src/entity/stripe/stripe-payment-intent.ts
+++ b/src/entity/stripe/stripe-payment-intent.ts
@@ -15,33 +15,20 @@
  *  You should have received a copy of the GNU Affero General Public License
  *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-
-import {
-  Column, Entity, JoinColumn, ManyToOne, OneToMany, OneToOne,
-} from 'typeorm';
-import { Dinero } from 'dinero.js';
 import BaseEntity from '../base-entity';
-import User from '../user/user';
+import { Column, Entity, JoinColumn, OneToMany, OneToOne } from 'typeorm';
+import StripePaymentIntentStatus from './stripe-payment-intent-status';
 import DineroTransformer from '../transformer/dinero-transformer';
-// eslint-disable-next-line import/no-cycle
-import StripeDepositStatus from './stripe-deposit-status';
-import Transfer from '../transactions/transfer';
+import { Dinero } from 'dinero.js';
+import StripeDeposit from './stripe-deposit';
 
 @Entity()
-export default class StripeDeposit extends BaseEntity {
-  @ManyToOne(() => User, { nullable: false, eager: true })
-  @JoinColumn()
-  public to: User;
-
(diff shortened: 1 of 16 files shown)


src/entity/stripe/stripe-payment-intent.ts
JustSamuel, line 1: NIT: prefer removing and creating a new file instead of renaming if they are unrelated to keep the git diff cleaner.

Yoronex: Hoe doe ik da?

But on a more serious note: the change seems somewhat weird, but this is also how I do the migration actually.

src/migrations/1722869409448-stripe-payment-intents.ts
```diff
@@ -0,0 +1,191 @@
...
+      new TableColumn({
+        name: 'transferId',
+        type: 'integer',
+        isNullable: true,
+      }),
+    );
+
+    const table = await queryRunner.getTable(this.DEPOSIT_TABLE_NAME);
+    await queryRunner.dropForeignKeys(this.DEPOSIT_TABLE_NAME, table.foreignKeys);
+    await queryRunner.dropTable(this.DEPOSIT_TABLE_NAME);
```
JustSamuel, line 188: I find this quite scary in production, also because this is slightly untested? 

Would it be an idea to add some test cases for this? that you take the suite for the stripe payment intents, revert the migration, add some legacy data, migrate, and run some tests on the new data as seed?

JustSamuel (changes requested):

src/migrations/1722869409448-stripe-payment-intents.ts
Yoronex: I do not actually delete the deposit table. First, I add the deposit-related columns back to the payment intent table and then I merge the two tables together into a single table. I drop the table here, because in the last step, I rename the payment intents table back to the deposit table. I have verified with a locally seeded SQLite database that the migrations work correctly.

Furthermore, I also do not really know how to build a robust test for this. The codebase will evolve and more attributes might be added to these entities. Furthermore, we are going to add more migrations. So how can I build a test case that reverts this single migration, even when we have more migrations and even migrations that change payment intents? And even if we can do that, these tests will probably break, because we write the code for the entities at this point in time, even though they might (and probably will) change.

JustSamuel: Fair, but then I do suggest to keep this in dev for another deployment cycle (so not include this in the next release) and see if it works in dev.

src/entity/stripe/stripe-payment-intent.ts
JustSamuel: Gewoon in je IDE een file verwijderen en een nieuwe aanmaken.

JustSamuel (approved):

PR #259, opened 2024-08-13
refactor(invoices): Invoices are now considered as a balance top-up

JustSamuel:
# Description
This PR is the second part of what is (hopefully) the last significant refactor of how invoices work in SudoSOS.

Currently, invoices are treated as an agreement that a debtor will pay the seller via bank transfer. As a result, when an invoice is created, **the invoiced amount is subtracted from the seller's balance and added to the buyers balance**.

Initially, this approach worked fine. However, as SudoSOS matured, more and more conflicts arose from this behavior. Coupled with the (ab)use of credit invoices and the ambition to create automated invoices, it was decided to treat invoices as a top-up. That is, **the balance of the invoiced user increases by the invoiced amount, while the seller's balance remains unaffected**.

This change means that, once this PR is migrated, invoiced transactions *will* appear in the sales overview, as they are simply sales. The seller should not need to concern themselves with whether a transaction is invoiced since the invoice is just a way for the buyer to top up their account. Consequently, we can start considering invoices as top-ups, both in SudoSOS and in the bookkeeping.

Adding the invoice to the bookkeeping is straightforward. The invoiced transactions are now included in the monthly sales overviews, meaning that VAT will be handled just like the rest of the transactions. The BACPM will *not* include the invoice in the bookkeeping; instead, a "TopupRequest"-PDF will be created and inserted into the bookkeeping. This top-up request is what the buyer will pay to the BACPM and is how the invoice gets paid off.

The reason for the "TopupRequest" instead of the invoice PDF is that the VAT of the invoice is already included in the monthly sales. Creating another PDF with even more VAT would be confusing. This also simplifies custom entries, as these will be included in the "TopupRequest" on a row by row basis and WILL be VAT-liable, whilst the invoiced amount will be represented as "SudoSOS balance, 0% vat". This makes it extremely clear what part of the invoice is just a "top-up" and what part is actually paying for goods and services outside of SudoSOS.

Credit invoices will be refactored to represent what they actually are: Payout requests by sellers linked to specific transactions. To this end, we will create another PR that will split the payout into a UserPayout and SellerPayout. The SellerPayout will specify what timeframe created this amount, allowing the BACPM to quickly generate a sales overview.

In conclusion, this PR decreases the complexity of invoices and makes the process more sustainable in the long run.

**TODO**
- [x] Create and test the migration
- [x] Create a new Sales Report
- [x] Create "TopupRequest"-pdf (or possibly diffeer
 
## Types of changes


src/controller/response/invoice-response.ts
```diff
@@ -22,6 +22,7 @@ import { BaseUserResponse } from './user-response';
 import { TransferResponse } from './transfer-response';
 import { PaginationResult } from '../../helpers/pagination';
 import { InvoiceState } from '../../entity/invoices/invoice-status';
+import { DineroObjectResponse } from './dinero-response';
 
 /**
  * @typedef {object} InvoiceStatusResponse
```
Yoronex, line 28: This typedef name does not match the name of the object below.

src/service/invoice-service.ts
```diff
@@ -137,39 +164,48 @@ export default class InvoiceService {
...
-    : InvoiceResponse {
+  public async asInvoiceResponse(invoice: Invoice)
+    : Promise<InvoiceResponse> {
+    const customEntries = invoice.invoiceEntries.map(InvoiceService.asInvoiceEntryResponse);
+    const relations: FindOptionsRelations<SubTransactionRow> = {
+      product: {
+        vat: true,
+      },
+    };
+    const subTransactionRows = await this.manager.find(SubTransactionRow, { where: { invoice: { id: invoice.id } }, relations });
```
Yoronex, line 188: Why don't you just create an inverse relation from an `Invoice` to its `SubTransactionRows`? That saves this (in my opinion) pretty ugly query in a response builder.

src/service/invoice-service.ts
```diff
@@ -137,39 +164,48 @@ export default class InvoiceService {
...
-      ...this.asBaseInvoiceResponse(invoice),
-      invoiceEntries: invoice.invoiceEntries
-        ? invoice.invoiceEntries.map(this.asInvoiceEntryResponse) : [],
+      ...InvoiceService.asBaseInvoiceResponse(invoice),
+      invoiceEntries: [...customEntries, ...transactionEntries],
     } as InvoiceResponse;
   }
 
-  public static toResponse(invoices: Invoice | Invoice[], entries: boolean): BaseInvoiceResponse | InvoiceResponse | BaseInvoiceResponse[] | InvoiceResponse[] {
+  public async toResponse(invoices: Invoice | Invoice[], entries: boolean): Promise<BaseInvoiceResponse | InvoiceResponse | BaseInvoiceResponse[] | InvoiceResponse[]> {
```
Yoronex, line 197: I am really not a fan of one method possibly returning four different response types. It is possible to separate all these in their own methods? With so many responses, implicit typing is no longer a thing.

src/service/invoice-service.ts
```diff
@@ -428,23 +356,25 @@ export default class InvoiceService {
...
+   */
+  async removeSubTransactionInvoice(invoice: Invoice) {
+    const subTransactionRows = await this.manager.find(SubTransactionRow, { where: { invoice: { id: invoice.id } } });
+    const promises: Promise<any>[] = [];
+    subTransactionRows.forEach((tSubRow) => {
+      const row = tSubRow;
+      row.invoice = null;
+      promises.push(this.manager.save(SubTransactionRow, row));
+    });
+    return Promise.all(promises);
```
Yoronex, line 371: Can't you just replace this complete function with an `onDelete: 'SET NULL'` on the `SubTransactionRow` side? Then, this will all be done automatically by the database (and you never make a mistake).

src/service/invoice-service.ts
```diff
@@ -500,82 +417,75 @@ export default class InvoiceService {
...
+            vat: true,
+          },
+        },
+        container: true,
+      },
+      pointOfSale: true,
+    };
+
+    const transactions = await this.manager.find(Transaction, { where: { id: In(tIds) },
+      relations });
```
Yoronex, line 460: Is there no way we can do all this with the existing `TransactionService.getTransactions` method? Because this feels very redundant.

src/controller/response/invoice-response.ts
JustSamuel: ?

src/service/invoice-service.ts
JustSamuel: No, getTransactions returns a paginated base response which does not allow the user to extract invoice information.

The transaction service also needs the getOptions refactor, sadly.

src/entity/invoices/invoice.ts
```diff
@@ -64,16 +65,13 @@ export default class Invoice extends BaseEntity {
...
-  @OneToOne(() => InvoiceStatus, { nullable: true, eager: false })
-  @JoinColumn({ name: 'latestStatusId' })
-  public latestStatus?: InvoiceStatus;
+  // TODO drop during migration
+  // /**
+  //  * The current status of the invoice
+  //  */
+  // @OneToOne(() => InvoiceStatus, { nullable: true, eager: false })
+  // @JoinColumn({ name: 'latestStatusId' })
+  // public latestStatus?: InvoiceStatus;
```
Yoronex, line 74: When will this be migrated?

src/entity/invoices/invoice.ts
```diff
@@ -156,6 +163,9 @@ export default class Invoice extends BaseEntity {
   })
   public date: Date;
 
+  @OneToMany(() => SubTransactionRow, (row) => row.invoice, { cascade: true })
+  public subTransactionRows: SubTransactionRow[];
+
```
Yoronex, line 168: Wait... Cascade? So if the subtransactions are deleted, the invoice will be deleted too?

src/entity/transactions/sub-transaction-row.ts
```diff
@@ -41,7 +41,7 @@ export default class SubTransactionRow extends BaseEntity {
   })
   public amount: number;
 
-  @ManyToOne(() => Invoice, { nullable: true })
+  @ManyToOne(() => Invoice, { nullable: true, onDelete: 'CASCADE' })
```
Yoronex, line 44: Same here... If the invoice is deleted, this subtransactionrow will be deleted too? I would say you would use `RESTRICT`, because you do not want to delete a subtransactionrow that you have invoiced

src/service/invoice-service.ts
```diff
@@ -88,9 +97,11 @@ export function parseInvoiceFilterParameters(req: RequestWithToken): InvoiceFilt
   };
 }
 
+// TODO: delete the 'balanced' invariant from invoices.
+
```
Yoronex, line 101: TODO

src/service/invoice-service.ts
```diff
@@ -106,9 +117,24 @@ export default class InvoiceService {
       amount: invoiceEntries.amount,
       priceInclVat: invoiceEntries.priceInclVat.toObject(),
       vatPercentage: invoiceEntries.vatPercentage,
+      custom: true,
```
Yoronex, line 120: Why is `custom` always true? We no longer support custom invoice entries, right?

src/service/invoice-service.ts
```diff
@@ -137,97 +165,72 @@ export default class InvoiceService {
       transfer: invoice.transfer ? TransferService.asTransferResponse(invoice.transfer) : undefined,
       description: invoice.description,
       pdf: invoice.pdf ? invoice.pdf.downloadName : undefined,
-      currentState: InvoiceService.asInvoiceStatusResponse(invoice.latestStatus),
+      currentState: InvoiceService.asInvoiceStatusResponse(invoice.invoiceStatus[invoice.invoiceStatus.length - 1]),
```
Yoronex, line 168: I'm not sure whether we should assume that the last invoice status in the list is the most recent one. Maybe write a small lambda function to pick the newest?

Yoronex (approved):

src/service/invoice-service.ts
JustSamuel: I will remove them later, as this would blow up the scope of this PR

(requires changing seeder, pdf gen, etc.)

src/service/invoice-service.ts
JustSamuel: fixed

Issue #244, opened 2024-08-01
[Feature] Nested product categories


Yoronex:
### What would you like?

It would be great if we can further classify products into more specific categories, like "Alcoholic" - "Beers" - "Craft beers" - "Blond". 

### Why is this needed?

Better filtering of products based on their category in the POS. Also works amazing for statistics!

### How could it be implemented?

Make `ProductCategory` a `TreeEntity` with a closure table, because it is necessary to efficiently query a category's root (because that determines in which tab the product is shown in the POS). https://orkhan.gitbook.io/typeorm/docs/tree-entities#closure-table

### Other information

Dependent on https://github.com/GEWIS/sudosos-backend/pull/233.

PR #261, opened 2024-08-14
feat: nested product categories

Yoronex:
# Description

Upgrades product categories from simple, standalone entities to tree entities. This means that each category can now optionally have a parent category, which allows us to further distinguish different types of products. Examples are Wines (white, red, rose) and Beers (pils, tripel, blond, dubbel, etc). The `GET /productcategories` endpoint now has optional query filters to only get root categories (which are the tabs in the POS) or to only get leaf categories. However, for the POS to understand in which root category a subcategory belongs, its best for the POS to simply request all categories from the backend and then create a local mapping of categories to their root category.

Other uses are mostly statistics-related, so we can for example better understand what types of products are most preferred by users.

Fixes https://github.com/GEWIS/sudosos-backend/issues/244.

## Types of changes

### DIFF
--- a/src/controller/product-category-controller.ts
+++ b/src/controller/product-category-controller.ts
@@ -74,6 +74,7 @@ export default class ProductCategoryController extends BaseController {
    * @operationId getAllProductCategories
    * @tags productCategories - Operations of productcategory controller
    * @security JWT
+   * @param {boolean} onlyRoot.query - Whether to return only root categories
    * @param {integer} take.query - How many product categories the endpoint should return
    * @param {integer} skip.query - How many product categories should be skipped (for pagination)
    * @return {PaginatedProductCategoryResponse} 200 - All existing productcategories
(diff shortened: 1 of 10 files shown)


src/controller/product-category-controller.ts
```diff
@@ -74,6 +74,8 @@ export default class ProductCategoryController extends BaseController {
    * @operationId getAllProductCategories
    * @tags productCategories - Operations of productcategory controller
    * @security JWT
+   * @param {boolean} onlyRoot.query - Whether to return only root categories
+   * @param {boolean} onlyLeaf.query - Whether to return only leaf categories
    * @param {integer} take.query - How many product categories the endpoint should return
    * @param {integer} skip.query - How many product categories should be skipped (for pagination)
    * @return {PaginatedProductCategoryResponse} 200 - All existing productcategories
```
JustSamuel, line 81: NIT: Could you remove the pagination from the response? In almost all cases you want all categories, and this just complicates front end code.

JustSamuel (approved):
Are you sure this is not a breaking change? I.e, will the current POS and Dashboard work with this? 

Otherwise it would force an update of the dashboard and pos before we can merge this into main

Yoronex:
As long as you do not create any subcategories before these are properly implemented, nothing will break.

src/controller/product-category-controller.ts
Yoronex: This will make this PR a breaking change, if that is what you want?

Issue #260, opened 2024-08-14
[Refactor] Major invoice changes: view invoices as deposits and replace credit invoices with SellerPayouts


Yoronex:
### What would you like?

Over the last few days/weeks, multiple issues with the existing invoice implementation came to light, like https://github.com/GEWIS/sudosos-backend/issues/255, https://github.com/GEWIS/sudosos-backend/issues/249. Furthermore, automatic invoices (https://github.com/GEWIS/sudosos-backend/issues/197) were one of the desired features to implement over the summer, yet they were not so trivial to implement (https://github.com/GEWIS/sudosos-backend/issues/197#issuecomment-2258780616). Even then, the invoice implementation already started to crack.

To clearly distinguish SudoSOS balance from BAC bank balance, hereafter I will use "SudoSOS Coin" as the currency of SudoSOS.

## Problem

All these issues with invoices boil down to two major design flaws:
1. Even though it was one of the design philosophies of SudoSOS, SudoSOS Coin is not always transfered between two users. For transactions this is the case: the user trades SudoSOS Coin for physical goods. However, to get SudoSOS Coin, the user has to spend EUR. These euros are send to the BAC bank account and in exchange, the user gets SudoSOS Coin. You would expect that the SudoSOS Coin is then taken from the BAC and placed into the user's account, but this is not the case. With the current implementation, these SudoSOS Coins are taken from the void, i.e. created out of nothing. All transfers behave this way (as a SudoSOS Coin exchange from/to the void), except for one, namely _invoices_. These are always an exchange of SudoSOS Coin between two users. This creates an edge case that has to be handled correctly in for example the transaction reports.
2. "Credit" invoices are not actually credit invoices. They are used as an overview of all sold products within a time frame, which the BAC uses to get SudoSOS Coin out of SudoSOS for sellers. For these "invoices" another edge case is created, namely that "credit" invoices also only exchange with the void (and **not** with another user like with regular invoices). This causes problems https://github.com/GEWIS/sudosos-backend/issues/249#issuecomment-2282294319.

The whole thing is that invoices are handled as transactions out of SudoSOS, even though they are registered within SudoSOS. The idea was that the sold products are paid for via the BAC bank account and not SudoSOS, so they should not be included in the SudoSOS overviews. However, this edge case adds a lot of complexity.

## Solution

[...]

PR #266, opened 2024-08-22
feat: seller payouts

Yoronex:
# Description

The `SellerPayouts` part of https://github.com/GEWIS/sudosos-backend/issues/260. This PR adds the new entity with all related endpoints and a migration to replace "credit" invoices with these new seller payouts, but there are some things to watch out for.
1. @JustSamuel mentioned to me personally that not all "credit" invoices can be migrated automatically, because they are not all correct. Before and after executing the migration, we should ensure that all invoice and seller payout data is correct.
2. The amount of each seller payout is now zero by default, because there is a dependency on https://github.com/GEWIS/sudosos-backend/pull/263 to calculate this correct amount.

Part of https://github.com/GEWIS/sudosos-backend/issues/260.

## Types of changes


JustSamuel:
Note that this is also dependent on #278

JustSamuel:
seller payouts should be returned desc. endDate

src/controller/debtor-controller.ts
```diff
@@ -314,17 +315,11 @@ export default class DebtorController extends BaseController {
...
+    let fromDate, toDate;
     try {
-      fromDate = asDate(req.query.fromDate);
-      toDate = asDate(req.query.toDate);
-      fromDate.setUTCHours(0, 0, 0, 0);
-      toDate.setUTCHours(0, 0, 0, 0);
-      if (toDate < fromDate) {
-        res.status(400).json('toDate must be after fromDate');
-        return;
-      }
```
JustSamuel, line 327: I suggest we fix this down stream in the sales report PR?

JustSamuel (approved):

Issue #268, opened 2024-08-23
[Bug] Email template missing in production build


Yoronex:
### Current behavior

![WhatsApp Image 2024-08-22 at 18 24 40](https://github.com/user-attachments/assets/22c2c252-241e-4f8b-bf3b-4c00b89b62a6)

HTML template seems to be missing from the production build. This probably happens because `tsc` does not include any non-TypeScript files.

### Steps to reproduce

Send an email and nothing will happen.

PR #269, opened 2024-08-23
fix: add mail templates to tsconfig

JustSamuel:
https://github.com/GEWIS/sudosos-backend/issues/268
## Types of changes

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -18,7 +18,8 @@
     },
     "include": [
         "src/**/*",
-        "test/**/*"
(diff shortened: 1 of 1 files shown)


CodeNamedRobin (approved):

Yoronex:
Unfortunately, this fix does not actually resolve the problem, as the template is still not included in the build. Therefore, I made a PR that creates a new "static" folder where we can put these non-TypeScript files. https://github.com/GEWIS/sudosos-backend/pull/270

Issue #176, opened 2024-04-30
Improved migration and integration testing


JustSamuel:
So, as I made a mistake in my database migration (#175) I propose a better flow. 

For each migration we should create a test case as follows:
- Create schema from before database changes.
- Run synchronize after database changes
- [Use TypeORM automigration creator](https://orkhan.gitbook.io/typeorm/docs/migrations#generating-migrations)
- Use that to rollback in the test env
- Run our own migrations
- Run tests cases

Of course this only works on MySQL as you need a persistent DB schema across the test cases. We already have a test similar to that, but we use our own `up` and `down` defined migrations which means that the precondition of having the database as it was before we changed the database is not met.

This should find those mistakes

Note: why don't we just always use the automigrate? I encourage you to check out the ungodly and unmanage sql that TypeORM spits out. I think we are better of writing our own, sensible, migrations.

JustSamuel:
Going to close this since the automigration rollback wouldn't really work since they also hardcode all the keys and relations. This makes creating a function rollback quite difficult. My suggestion would just be good testing? Keep it running on develop for some time, that worked.

Better yet, try and prevent migrations.

PR #272, opened 2024-08-23
fix: migration mistakes regarding nested product categories

Yoronex:
# Description

Fixes two mistakes/errors in the nested product categories migration:
1. Even though a [closure table](https://orkhan.gitbook.io/typeorm/docs/tree-entities#closure-table) is used to create the tree structure for product categories, TypeORM still creates a `parentId` column in the `product_category` table. After some thinking, this makes sense: how else would you distinguish your parents from your ancestors? Stupid me was stupid.
2. For every record in the `product_category` table, the closure table should include a mapping from each record to itself. The migration did not do this (but now it does)!

These two things therefore will need manual fixing in the test environment, as there already is a newer migration in the list. However, we can also undo the Stripe migration and then redo this migration as well, as the Stripe migration is not destructive.

## Types of changes

### DIFF
--- a/src/migrations/1722517212441-nested-product-categories.ts
+++ b/src/migrations/1722517212441-nested-product-categories.ts
@@ -15,10 +15,20 @@
  *  You should have received a copy of the GNU Affero General Public License
  *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-import { MigrationInterface, QueryRunner, Table, TableForeignKey } from 'typeorm';
+import { MigrationInterface, QueryRunner, Table, TableColumn, TableForeignKey } from 'typeorm';
 
 export class NestedProductCategories1722517212441 implements MigrationInterface {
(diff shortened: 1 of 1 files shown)


JustSamuel:
Ugh, another migration issue :(

I guess it might be time to create a testing setup where the migration is actually tested? I had something in mind, https://github.com/GEWIS/sudosos-backend/issues/176, but the issue was that you need a "previous" database to migrate and test from.

My suggestion, is we make a ci job that does the following:

pull and build sudosos from the _target_ branch, sync the mariadb and then go to the source version of the PR so that we can actually migrate and test?

JustSamuel (approved):

Yoronex:
I think we need to make an "initial" migration to build upon. Then you can build your complete database schema by using the migrations and compare this to the result of a TypeORM synchronize. Any discrepancies you can immediately detect. However, making such an "initial" migration is going to be a _huge_ pain...

PR #279, opened 2024-08-24
Feature/seller payouts pdf

JustSamuel:
# Description
This PR is actually a combination of #263, #266 and #278 

## Types of changes


JustSamuel:
@Yoronex maybe you can find why the mariadb suite hangs? it must be a "after" since otherwise the test case would time out.

src/entity/transactions/payout/payout-request.ts
```diff
@@ -23,13 +23,14 @@ import User from '../../user/user';
...
 import PayoutRequestPdf from '../../file/payout-request-pdf';
-import { hashJSON } from '../../../helpers/hash';
-import PayoutRequestPdfService from '../../../service/payout-request-pdf-service';
 import BasePayout from './base-payout';
+import { PdfAble } from '../../file/pdf-able';
+import PayoutRequestPdfService from '../../../service/pdf/payout-request-pdf-service';
+import { PAYOUT_REQUEST_PDF_LOCATION } from '../../../files/storage';
 
 @Entity()
 // TODO Migration
```
Yoronex, line 32: Officially not part of this PR, but if this TODO is resolved, please remove it here as well :)

src/service/pdf/seller-payout-pdf-service.ts
```diff
@@ -0,0 +1,70 @@
...
+  async getParameters(entity: SellerPayout): Promise<SellerPayoutParameters> {
+    const { amount, startDate, endDate, reference } = entity;
+    const report = await new SalesReportService().getReport({
+      fromDate: startDate,
+      tillDate: endDate,
+      forId: entity.requestedBy.id,
+    });
+
+    // TODO? throw error if amount and report amount are different?
+    console.error(amount.getAmount(), report.totalInclVat.getAmount());
```
Yoronex, line 49: `console.error`? Seems like a debugging line?

src/service/pdf/seller-payout-pdf-service.ts
```diff
@@ -0,0 +1,70 @@
...
+    assert(amount.getAmount() === report.totalInclVat.getAmount(), 'Amounts do not match');
```
Yoronex, line 50: We had this discussion before somewhere, but I think we should not disallow the user to download payout reports with an invalid amount, as viewing the report is the only way to know what the payout amount should be (by design).

test/unit/service/seller-payout-service.ts
```diff
@@ -337,6 +337,7 @@ describe('SellerPayoutService', () => {
       expect(sellerPayout.transfer.fromId).to.equal(params.requestedById);
       expect(sellerPayout.transfer.toId).to.be.null;
       const transferCreationDate = Math.round(sellerPayout.transfer.createdAt.getTime() / 1000);
+      console.error(actualEndDate, transferCreationDate, sellerPayout.endDate, sellerPayout.transfer.createdAt);
```
Yoronex, line 340: `console.error`?

Yoronex (changes requested):
There is one major thing missing, namely an actual HTTP endpoint to download such a PDF file. An endpoint is already added, but it needs an implementation. See https://github.com/GEWIS/sudosos-backend/blob/e7f5a6ad810672bba3e73cb4af747566e310b387/src/controller/seller-payout-controller.ts#L213.

Yoronex (approved):

PR #225, opened 2024-07-25
feat: Write-Off's

JustSamuel:
# Description
This PR adds Write Offs to SudoSOS. A `WriteOff` is the act of closing an account which has a negative balance by giving it a "top-up" of this negative amount. The use case is that some accounts are left with a negligible debt which is easier to write-off than to try and get your money back. 

The idea for the larger debts might be to create a special type of top-up which the user can pay via a link and a topup instead of having to log in again. Easier to send someone a SudoSOS "tikkie" than have them log in again.

TODO
- [x] tests
- [x] seeder
- [ ] pdf''s (seperate PR)

*NB: This PR is rebased on the transfer-vat rework.*

## Types of changes
- New feature _(non-breaking change which adds functionality)_

PR #291, opened 2024-08-30
[DO NOT MERGE] fix: hanging test suite

JustSamuel:
# Description
Test suite seems to hang in the CI, this branch / PR is to test and find out what causes it.

Caused in PR #279 

Possible targets:
- [ ] `debtor-controller.ts`
- [ ] `payout-request-controller.ts`
- [ ] `seller-payout-controller.ts`
- [ ] `user-controller.ts`
- [ ] `invoice-pdf-service.ts`
- [ ] `payout-request-pdf-service.ts`
- [ ] `report-pdf-service.ts`
- [ ] `report-service.ts` 
- [ ] `payout-request-service.ts`
- [ ] `report-pdf-service.ts`
- [ ] `seller-payout-service.ts`

## Types of changes


JustSamuel:
I tried a lot of combinations of different test suits, but this doesn't seem to actually make a difference? Any _n_ test cases seem to cause it to hang, where _n_ is around 2 or 3.

Using `wtfnode` gives me the impression that we are failing to close a typeorm session somewhere, which makes sense since it doesn't fail on sqlite (which has no connection pooling).

If it was a bad test case with a weird hanging promise, then sqlite should have had the same problem.

JustSamuel:
Nvm, this seems to have been an issue since a looooongggg time. At least #225 but could have been longer. I suggest to just try and find the hanging mariadb connection instead of focusing on which test. Most likely it is not related to a test case but to our before and after of the test cases.

Issue #280, opened 2024-08-25
[Feature] Maintenance mode


JustSamuel:
### What would you like?

Disable SudoSOS for all non admins as a toggle 

### Why is this needed?

I don’t want people to mess with SudoSOS during migrations

### How could it be implemented?

Add a “maintenance mode” server setting, a middle ware to check and block request if enabled and a endpoint to toggle

PR #295, opened 2024-09-01
feat: add maintenance mode to restrict access to SudoSOS during upgrades

Yoronex:
# Description

Adds maintenance mode to the `restriction-middleware`. The following rules apply:
- Maintenance mode is added as a server setting
- All endpoints (except LDAP login, local login and `/ping`) will return a 503 once maintenance mode is enabled.
- All users that have the permission `Maintenance override` can always access all endpoints, even during maintenance.
- Maintenance mode can be enabled/disabled using a new endpoint.

Fixes https://github.com/GEWIS/sudosos-backend/issues/280.

## Types of changes

### DIFF
--- a/src/controller/authentication-controller.ts
+++ b/src/controller/authentication-controller.ts
@@ -118,6 +118,7 @@ export default class AuthenticationController extends BaseController {
           body: { modelName: 'AuthenticationLocalRequest' },
           policy: async () => true,
           handler: this.LocalLogin.bind(this),
+          restrictions: { availableDuringMaintenance: true },
         },
         PUT: {
           body: { modelName: 'AuthenticationResetTokenRequest' },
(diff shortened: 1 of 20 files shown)


src/controller/authentication-controller.ts
```diff
@@ -150,7 +151,7 @@ export default class AuthenticationController extends BaseController {
     const body = req.body as AuthenticationMockRequest;
 
     // Only allow in development setups
-    if (process.env.NODE_ENV !== 'development') return false;
+    if (process.env.NODE_ENV !== 'development' && process.env.NODE_ENV !== 'test') return false;
```
JustSamuel, line 154: Won't this always be false

JustSamuel (approved):

src/controller/authentication-controller.ts
Yoronex: No, because when you are in a `production` environment, you cannot perform a mock login. In `development` (your local environment) and `test` (when running test suites) this endpoint should be enabled.

PR #296, opened 2024-09-01
refactor: change `UserType` enum from integers to strings

Yoronex:
# Description

Right now, we store the `UserType` as an integer in the database. However, we communicate the full name (string) to the frontend. To be more consistent, the `UserType` in the database are now strings as well. This should also fix any issues where we expect an UserType integer as input.

## Types of changes


JustSamuel:
Not a chore but a refactor

src/database/migrate.ts
```diff
@@ -36,11 +36,15 @@ export default async function migrate() {
...
+    application.logger.debug('Synchronize...');
     await application.connection.synchronize();
+    application.logger.debug('Fake migrations...');
     await application.connection.runMigrations({ transaction: 'all', fake: true });
+    application.logger.debug('Revert last migration...');
     await application.connection.undoLastMigration({ transaction: 'all' });
+    application.logger.debug('Run last migration...');
     await application.connection.runMigrations({ transaction: 'all' });
-    await application.connection.close();
+    await application.connection.destroy();
```
JustSamuel, line 47: Is dit waarom hij bleef hangen????

JustSamuel (approved):

src/database/migrate.ts
JustSamuel: image
het lijkt iets te doen

PR #297, opened 2024-09-01
refactor(invoice): remove custom invoice entries

JustSamuel:
# Description
Final part of the Invoicing rework, this removes the ability to add custom entries on invoices.

TODO
- [x] test the deletion of invoices moving entries

## Types of changes


src/entity/invoices/invoice.ts
```diff
@@ -160,6 +150,10 @@ export default class Invoice extends PdfAble(BaseEntity) {
   @OneToMany(() => SubTransactionRow, (row) => row.invoice, { cascade: false })
   public subTransactionRows: SubTransactionRow[];
 
+  @ManyToMany(() => SubTransactionRow, { cascade: false })
+  @JoinTable()
+  public subTransactionRowsDeletedInvoice: SubTransactionRow[];
```
Yoronex, line 155: Maybe it is better to make this a separate entity? Because there are multiple problems with simply using a many-to-many table:
1. You cannot delete the same sub transaction row from the same invoice multiple times (if you are truly dumb and need to do this).
2. You don't know when you deleted a certain sub transaction row, so it is not clear what the invoice was at which point in time.

src/service/invoice-service.ts
```diff
@@ -205,27 +186,23 @@ export default class InvoiceService {
...
+      return acc + t.value.amount;
+    }, 0);
 
-    if (baseTransactions.length !== 0) {
-      baseTransactions.forEach((t) => {
-        if (t.from.id !== forId) throw new Error(`Transaction from ${t.from.id} not from user ${forId}`);
-        dineroObjectRequest.amount += t.value.amount;
-      });
-    }
+    // TODO: Should we error if the sum is not equal to the amount?
```
Yoronex, line 201: No, I don't think we should. Like with the `SellerPayout`, I think it is best to just make it very clear in the frontend that something does not add up. Then, you can fix it afterwards. Creating invoices is a reversable action, right?

test/unit/service/invoice-pdf-service.ts
```diff
@@ -235,93 +237,7 @@ describe('InvoicePdfService', async (): Promise<void> => {
...
-        description: 'Product unsupported VAT',
-        amount: 1,
-        priceInclVat: DineroTransformer.Instance.from(500),
-        vatPercentage: 5,
-      } as InvoiceEntry];
-
-      expect(() => invoice.pdfService.entriesToProductsPricing(invoice)).to.Throw('Unsupported vat percentage 5 during pdf generation.');
-    });
-  });
+  // TODO: test invoiceToPricing and subTransactionRowToProduct
```
Yoronex, line 240: TODO

Yoronex (changes requested):

test/unit/service/invoice-pdf-service.ts
Yoronex: TODO is still here :(

Yoronex (approved):

PR #303, opened 2024-09-04
fix: log4js logging

JustSamuel:
# Description
Should fix the log4js workers not all printing to console.

[Found here](https://github.com/log4js-node/log4js-node/issues/265#issuecomment-544974438)

## Types of changes


tomudding (approved):
Glad to see that this works! 🚀

PR #305, opened 2024-09-05
[DRAFT] chore: update dependencies

Yoronex:
# Description

Updates all dependencies to the latest versions. After merging this, we should consider enabling Dependabot for this repository, as individually updating dependencies is easier than updating everything at once.

There is however one problem I encountered and was not able to solve, namely updating `chai`, the package we use for all our unit tests. Since version 5, `chai` does only support Javascript modules and no longer the classic `commonjs`. However, this seems to cause problems with the Typescript compiler I was not able to solve. I encountered at least one of the following problems:
- `ts-node` was not able to read `.ts` module files when I set `"type": "module"` in `package.json`.
- TypeORM could not longer imply the column types from the entity attribute types. This happened when I replaced `ts-node` with `tsx`.

@JustSamuel do you perhaps have time to look at this problem? All dependencies are currently up to date, except all `chai`-related packages (including `@types`).

## Types of changes


Yoronex:
When starting the project using the `tsx` Typescript compiler (`watch`, `seed`, `test`, etc), TypeORM was unable to determine the column types.

JustSamuel:
`tsx` will never work with `TypeOrm` unless you do kinky stuff:

https://github.com/evanw/esbuild/issues/257

JustSamuel:
I say give up and let this be?

Yoronex:
But we will have to update our dependencies at some point. Right now, `chai` is not up to date, which causes a whole bunch of other packages to also not be up to date. `mocha` still works with the old version, but I have no clue how long that will remain the case...

Yoronex:
Okay, I suggest we merge this PR and open a new issue for updating `chai`. Then we can at least have the existing packages be up to date.

JustSamuel:
Sure, but what reason do we have then to switch to nodemon?

Yoronex:
We need to use `nodemon` because `ts-node` is unable to watch any files. `nodemon` automatically reloads the application once a file has changed.

JustSamuel:
> We need to use `nodemon` because `ts-node` is unable to watch any files. `nodemon` automatically reloads the application once a file has changed.

Yeah, but we used to use ts-node-dev for that, why switch?

JustSamuel (approved):

Yoronex:
https://www.npmjs.com/package/ts-node-dev ts-node-dev has had its last update almost 2,5 years ago. In the repository's issues, people are actively doubting whether this package is still being maintained. As `ts-node` is still being developed, it is undesirable to stick to an old version just to have the active reload functionality.

nodemon is actively being developed and - together with ts-node - can do the exact same thing as ts-node-dev. Both packages also see far more downloads/users, so I have more faith in these two packages being maintained than ts-node-dev.

Does this answer your question/concerns?

PR #332, opened 2024-09-18
fix: handle ObjectGUID as Buffer correctly

JustSamuel:
# Description
Mixing little endian and big endian should be considered a warcrime. 
https://joelitechlife.ca/2021/04/05/get-objectguid-of-microsoft-active-directory-using-ldapsearch/

## Types of changes

### DIFF
--- a/src/helpers/ad.ts
+++ b/src/helpers/ad.ts
@@ -68,12 +68,10 @@ export interface LDAPUser {
  */
 export async function bindUser(manager: EntityManager,
   ADUser: { objectGUID: Buffer }, user: User): Promise<LDAPAuthenticator> {
-  const auth = manager.create(LDAPAuthenticator, {
+  return manager.save(LDAPAuthenticator, {
     user: user,
     UUID: ADUser.objectGUID,
   });
-  await manager.save(auth);
-  return auth;
 }
 
 /**
--- a/src/helpers/ad.ts
+++ b/src/helpers/ad.ts
@@ -38,7 +38,7 @@ export function getLDAPSettings() {
 }
 
 export interface LDAPResponse {
-  objectGUID: string,
+  objectGUID: Buffer,
   whenChanged: string,
 }
 
@@ -54,7 +54,7 @@ export interface LDAPUser {
   memberOfFlattened: string[],
   givenName: string,
   sn: string,
-  objectGUID: string,
+  objectGUID: Buffer,
   mail: string,
   mNumber: number | undefined;
 }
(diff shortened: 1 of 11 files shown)


JustSamuel:
Note @Yoronex that this is quite a nifty migration, so this should only be done under supervision so that we can recover it if needed.

src/service/authentication-service.ts
```diff
@@ -268,9 +268,10 @@ export default class AuthenticationService extends WithManager {
       const { searchEntries } = await client.search(ldapSettings.base, {
         scope: 'sub',
         filter: filterstr,
+        explicitBufferAttributes: ['objectGUID'],
       });
       if (searchEntries[0]) {
-        ADUser = userFromLDAP(searchEntries[0]);
+        ADUser = userFromLDAP(searchEntries[0] as any as LDAPResult);
```
Yoronex, line 274: Same here.

test/unit/service/ad-service.ts
```diff
@@ -308,6 +317,10 @@ describe('AD Service', (): void => {
         { where: { authenticateAs: { id: newOrgan.id } }, relations: ['user'] },
       )).map((mAuth) => mAuth.user.id);
 
+      console.error((await MemberAuthenticator.find(
+        { where: { authenticateAs: { id: newOrgan.id } }, relations: ['user'] },
+      )));
+      console.error(secondMembers);
```
Yoronex, line 323: `console.error`

src/service/authentication-service.ts
JustSamuel: This is directly from the `client.search`, so no.

src/service/ad-service.ts
```diff
@@ -112,7 +114,7 @@ export default class ADService extends WithManager {
       // Extract members
       const shared = sharedAccounts[i];
       const result = await this.getLDAPGroupMembers(client, shared.dn);
-      const members: LDAPUser[] = result.searchEntries.map((u) => userFromLDAP(u));
+      const members: LDAPUser[] = result.searchEntries.map((u) => userFromLDAP(u as any as LDAPResult));
```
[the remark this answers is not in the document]
JustSamuel: fixed

src/helpers/ad.ts
```diff
@@ -67,11 +67,11 @@ export interface LDAPUser {
...
-  ADUser: { objectGUID: string }, user: User): Promise<LDAPAuthenticator> {
-  const auth = Object.assign(new LDAPAuthenticator(), {
-    user,
+  ADUser: { objectGUID: Buffer }, user: User): Promise<LDAPAuthenticator> {
+  const auth = manager.create(LDAPAuthenticator, {
+    user: user,
     UUID: ADUser.objectGUID,
-  }) as LDAPAuthenticator;
+  });
   await manager.save(auth);
```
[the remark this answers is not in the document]
JustSamuel: fixed

test/unit/service/ad-service.ts
JustSamuel: fixed

src/service/authentication-service.ts
Yoronex: Then why the need to cast it, if `searchEntries[0]` is already of type `any`?

JustSamuel: It's not, its of type Entry which is an object of key-value pairs with strings as key. They decided not to specify any of the keys, not even the default AD ones which we use.

JustSamuel: A possible refactor would be to have a helper function which turns the Entry into an LDAPResult and even type checks it for sanity, however I do think this is out-of-scope of this PR. Previously before this PR the any to type conversion just happened in the `userFromLDAP` function itself, so nothing changed.

Yoronex (approved):

Issue #330, opened 2024-09-18
[Feature] Always sync AD service accounts, even if they do not yet exist in SudoSOS


Yoronex:
### What would you like?

SudoSOS should always synchronize service accounts from AD.

### Why is this needed?

Service accounts can be used to integrate SudoSOS with external services, like Aurora or barcommissie.nl. You want to use service accounts, as they can only access the things they are allowed to. However, there is currently no way to use such accounts. SudoSOS only synchronizes AD accounts that exist in SudoSOS. However, for an AD user to be created, the user has to log in with AD once, but this is impossible for service accounts (which often do not have a password). 

### How could it be implemented?

SudoSOS should create a new user account for an AD user during synchronization, if the to-be-added user is a service account.

JustSamuel:
Priority high? Also, I think this can be implemented in the LDAP sync that we created for the members, we only really need to define the service accounts groups as a sudosos role / permission.

Yoronex:
Yes this is pretty high priority, because I am currently using my own API key, which is pretty dangerous if you ask me. Before, I used yours.

PR #340, opened 2024-09-30
Refactor LDAP sync to sync service structure

JustSamuel:
# Description
This PR changes the syncing of users to a user-first syncing. This means that instead of having different services scattered throughout the codebase, we focus on having a single interface that can be implemented and used in the general user sync.

A `sync-service` should implement a `sync`, `fetch` and `down` function. It can also be decided to override the `guard` function if needed. The new syncing structure is as follows:

The `user-sync-service` is akin to a compound design pattern and takes an array of services to use. Each of these services implements the `sync` function, enabling the user to be synced with an external service. This `sync` has a result object which can indicate one of the following:
- The sync was skipped for the provided user (i.e. wrong user type)
- The sync errored (GEWISDB is down)
- The sync result, which is either true for a successful sync or false for a bad sync.

A bad sync can be the case that the user does no longer exist in LDAP or the GEWISDB.
After the user has been synced by all the services, the following happens:
- If none of the syncs errored, we either `down` a user if all services returned a bad sync, or else do nothing.

`down` is the function which should clean up or remove data related to the user entity. For the LDAP sync this will be removing the LDAPAuthentication entry, but for the GEWISDB sync this will mean deactivating the user in SudoSOS as no longer existing in the GEWISDB is a reason to be deactivated. This bring us to the following question:

**Currently `down` is only called when _all_ syncs are bad. Is this desired?**

Take the cases that a user becomes inactive and removed from AD. The LDAP sync will start to report bad syncs. This could be grounds to remove the user from the LDAP table. However only calling `down` when _all_ syncs report bad feels safer as it makes for a build in safeguard. This is exceptionally clear from the other case, where a user is removed from GEWISDB but still exists in  AD. This _is_ strange, and we would expect a user to also be removed from LDAP. However if they are still in AD, they could still "login", meaning that removing their whole SudoSOS account could potentially lead to a duplicate being created if they login with LDAP. 

Therefore I  am more inclined to go with the "all syncs should report bad"  option. In the end, everything will still be cleaned. It will just take a bit longer with the safer route.

#330 

## Types of changes


src/service/sync/sync-service.ts
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Guard determines whether the user should be synced using this sync service.
+   *
+   * Not passing the guard will result in the user being skipped.
+   * A skipped sync does not count as a failure.
+   *
+   * @param user The user to check.
+   * @returns {Promise<boolean>} True if the user should be synced, false otherwise.
+   */
+  guard(user: User): Promise<boolean> {
```
Yoronex, line 54: Can't this method be `protected`? You only seem to use it in the `up()` method.

src/service/sync/sync-service.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  /**
+   * Synchronizes the user data with the external data source.
+   *
+   * @param user The user to synchronize.
+   * @returns {Promise<boolean>} True if the user was synchronized, false otherwise.
+   */
+  abstract sync(user: User): Promise<boolean>;
```
Yoronex, line 84: Can't this method be `protected`? You probably only want to run it if the guard is satisfied, which is enforced by `up()`.

src/service/sync/ldap-sync-service.ts
```diff
@@ -0,0 +1,102 @@
...
+export default class LdapSyncService extends SyncService {
+
+  // We only sync organs and members.
+  targets = [UserType.ORGAN, UserType.MEMBER];
+
+  private readonly ldapClient: Client;
+
+  private readonly roleManager: RoleManager;
+
+  constructor(manager: EntityManager, ldapClient: Client, roleManager: RoleManager) {
```
Yoronex, line 38: Manager is optional in the base class `SyncService`, but is not optional in this derivative class.

src/service/sync/ldap-sync-service.ts
```diff
@@ -0,0 +1,102 @@
...
+   */
+  async fetch(): Promise<User[]> {
+    const adService = new ADService(this.manager);
+
+    // syncSharedAccounts also creates accounts for all the members of the shared groups.
+    // Is this something we want?
+    await adService.syncSharedAccounts();
+
+    // syncUserRoles also creates a new account of a user is assigned to a role but does not yet exist in sudosos.
+    // Is this something we want?
```
Yoronex, line 97: Yes, because service integration accounts.

src/service/sync/user-sync-service.ts
```diff
@@ -0,0 +1,93 @@
...
+  private services: SyncService[];
+
+  constructor(services: SyncService[]) {
+    super();
+    this.services = services;
+  }
+
+  async syncUsers() {
+    const userTypes = this.services.flatMap((s) => s.targets);
+    console.log('Syncing users of types', userTypes);
```
Yoronex, line 37: Please use `log4js` instead of `console.log()`.

src/service/sync/user-sync-service.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (result.error) {
+        console.error('Sync failed for user', user.id);
+        continue;
+      }
+
+      console.log('Synced user', user.id);
+      if (result.result === false) {
+        console.error('User is detached', user.id);
+        await this.down(user);
+      }
```
Yoronex, line 57: I think it is better if `sync()` (or `up()`) always returns the sync result or `undefined`/`null` if you have not synchronized anything. If syncing returns an error, I think it is nicer to handle this with a custom exception.

src/service/sync/user-sync-service.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    return syncResult;
+  }
+
+  async down(user: User): Promise<void> {
+    for (const service of this.services) {
+      try {
+        await service.down(user);
+      } catch (error) {
+        console.error(error);
```
Yoronex, line 81: `log4js`

src/service/sync/user-sync-service.ts
JustSamuel: Where would you like to catch it? simply in the user-sync-service?

src/service/sync/ldap-sync-service.ts
JustSamuel: Don't we want a more definitive split between bots and users? And the code as it is now won't create accounts for bots, since they do not have a m-number.

src/service/sync/user-sync-service.ts
JustSamuel: Also, I think I prefer an objects with
```
{
  skipped: boolean;
  result: booleanl
}
```

As giving implicit values to undefined | null | boolean does not make the code clearer, and javascript equality checks are wack in general. (falsly vs thruthy)

src/entity/user/user.ts
```diff
@@ -46,6 +46,7 @@ export enum UserType {
   LOCAL_ADMIN = 'LOCAL_ADMIN',
   INVOICE = 'INVOICE',
   POINT_OF_SALE = 'POINT_OF_SALE',
+  INTEGRATION = 'INTEGRATION',
```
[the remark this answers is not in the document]
JustSamuel: Not really, your desire for the INTEGRATION users is what kicked of the refactor, so it seemed fitting to include.

src/service/sync/user/ldap-sync-service.ts
```diff
@@ -0,0 +1,225 @@
...
+    // But we do return true to indicate that the user is "bound" to the LDAP
+    if (user.type === UserType.MEMBER) return true;
+
+    this.logger.trace(`Updating user ${user} from LDAP.`);
+    user.firstName = ldapUser.displayName;
+    user.lastName = '';
+    user.canGoIntoDebt = false;
+    user.acceptedToS = TermsOfServiceStatus.NOT_REQUIRED;
+    user.active = true;
+    await this.manager.save(user);
```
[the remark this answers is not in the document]
JustSamuel: LDAP users are matched on LDAP UUID, so if someone deletes an user in LDAP and creates a new one SudoSOS will have no way of matching them. Opinion?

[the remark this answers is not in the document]
Yoronex: I think it is fine to create a new one when the LDAP UUID has changed, but what if (due to a mistake by CBC) the LDAP user is not included during one or more syncs, and then reappears? Wouldn't be the first time everyone is discharged from every committee for example.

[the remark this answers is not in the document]
JustSamuel: `MEMBERS` are synced by both `LDAP` and `GEWISWEB`, so disappearing from the LDAP sync for a day isn't that much of an issue. If a user still has LDAP  and tries to login but has an unique UUID it wil also be matched against the mNumber. So that should also be fine.

The only real edge case if we allow LDAP accounts for local_users, which could be possible as having an gewis.nl account isnt strictly required for an LDAP account.

src/gewis/service/gewisdb-sync-service.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (!gewisUser) return;
+    
+    // We do not delete the GewisUser, for if we ever want to undo deletions and keep the link to the m-number.
+    // await this.manager.delete(GewisUser, gewisUser);
+
+    // Sync deleting a user is quite impactful, so we only do it if the setting is explicitly set.
+    if (!await GewisDBSyncService.getAllowDelete()) return;
+
+    const currentBalance = await new BalanceService().getBalance(entity.id);
+    await UserService.closeUser(entity.id, true).then(() => {
```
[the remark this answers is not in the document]
JustSamuel: Disagree; there is a large difference between a try-catch block and .then and .catching a promise. If you have the email one, you would want to .catch it but not await, so you will node some construction for that.

Yoronex (approved):

PR #342, opened 2024-10-02
Feature/inactive administrative cost

SvenMokveldje:
Implemented the feature to create, find, delete inactive administrative costs. Next to that you can check which users are eligible either to send a notification or send a fine. Lastly it has 2 functions to send an email to a array of users for either the notification or to fine them and send a notification.


SvenMokveldje:
Zie nu dat het 10 euro per jaar is, en miste ook nog een paar use cases. Zoals niet checken op de transfer van de vorige administrative cost.

PR #343, opened 2024-10-05
Feature/inactive administrative cost

SvenMokveldje:
Created the inactive administrative cost feature. BAC treasurer can now check who has been inactive at SudoSOS and either send them an notification and after 3 years fine them. This will fine them until they have a balance of 0. After which the BAC treasurer can disable those accounts. 

(Auto locking will be added once this is implemented)

You can check inactive users, notify them, handout the administrative costs, create the inactive administrative cost entity and delete it. There is no update function as you would never change the amount (this is fixed) so the only update would be to delete it and lastly get certain entities.


SvenMokveldje:
For the test fails: The first one, I do not understand why the post-miration-test tries to equal it to 20, as the default_pagination should be either that of process.env which is 20 or the default put in by Samuel which is 25.

The second one, does not fail on my machine (hihi), but as I take a id bigger than that should exist it should be equal to null. But maybe someone with more knowledge can take a crack at it.

src/middleware/restriction-middleware.ts
```diff
@@ -74,7 +74,7 @@ export default class RestrictionMiddleware {
         return;
       }
     } catch (e) {
-      getLogger('RestrictionMiddleware').error(e);
+      log4js.getLogger('RestrictionMiddleware').error(e);
```
JustSamuel, line 77: ?

src/service/debtor-service.ts
```diff
@@ -324,7 +324,7 @@ export default class DebtorService {
   }
 
   /**
-   * Send an email to all users with the given ID, notifying them that they will get fined a certain amount. The date
+   * Email all users with the given ID, notifying them that they will get fined a certain amount. The date
```
JustSamuel, line 327: why change

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,320 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ *  @license
+ */
+import WithManager from '../database/with-manager';
```
JustSamuel, line 20: space

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,320 @@
...
+    : Promise<UserToInactiveAdministrativeCostResponse[]> {
+    const { notification } = params;
+    const differenceDate = notification ? 2 : 3;
+
+    const users = await User.find();
+    const eligibleUsers: UserToInactiveAdministrativeCostResponse[] = [];
+
+    // go through all users and get their last transfer and transaction
+    for (let i = 0; i < users.length; i += 1) {
+      const user = users[i];
```
JustSamuel, line 117: prefer `for const of `

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,320 @@
...
+        where: { fromId: user.id },
+        relations: { inactiveAdministrativeCost: true },
+      }));
+
+      const lastTransfer = userTransfers.length == 0 ? null : userTransfers
+        .reduce((prev, curr) => (((prev.createdAt < curr.createdAt) && curr.inactiveAdministrativeCost == null) ? curr : prev));
+      const userTransactions = ((await Transaction.find({ relations: ['from'] }))
+        .filter((t) => t.from.id === user.id));
+      const lastTransaction = userTransactions.length == 0  ? null : userTransactions
+        .reduce((prev, curr) => (prev.createdAt < curr.createdAt ? curr : prev));
```
JustSamuel, line 134: what does this do

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,320 @@
...
+
+      if (lastTransfer != null) if (InactiveAdministrativeCostService.yearDifference(lastTransfer.createdAt) < differenceDate) {
```
JustSamuel, line 136: !==

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,320 @@
...
+  /**
+   * Checks which users are eligible for either a notification or a fine.
+   * @param params
+   */
+  public async checkInactiveUsers(params: InactiveAdministrativeCostFilterParameters)
+    : Promise<UserToInactiveAdministrativeCostResponse[]> {
+    const { notification } = params;
+    const differenceDate = notification ? 2 : 3;
+
+    const users = await User.find();
```
JustSamuel, line 112: uhhh, you want to loop over `all` users? why not write this using query builder? 
the set of people who will be found is very low, meaning that this seems a bit much.

https://orkhan.gitbook.io/typeorm/docs/select-query-builder

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,320 @@
...
+    // Get amount from transfer
+    const amount: DineroObjectRequest = inactiveAdministrativeCost.transfer.amountInclVat.toObject();
+
+    // We create an undo transfer that sends the money back to the person.
+    const undoTransfer: TransferRequest = {
+      amount,
+      description: 'Deletion of InactiveAdministrativeCost',
+      fromId: 0,
+      toId: inactiveAdministrativeCost.fromId,
+    };
```
JustSamuel, line 171: research shows that it is a must that you link this to the inactive administration entity. I would say a `deleteTransfer` or whatever I called it in invoices.

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,320 @@
...
+   */
+  public async createInactiveAdministrativeCost(inactiveAdministrativeCostRequest: CreateInactiveAdministrativeCostRequest)
+    : Promise<InactiveAdministrativeCost> {
+    const { forId } = inactiveAdministrativeCostRequest;
+
+    // Calculate reduction amount
+    const user = await this.manager.findOne(User, { where: { id: forId } });
+    const userBalance = await new BalanceService(this.manager).getBalance(forId);
+
+    const monetaryAmount = (userBalance.amount.amount < 5) ? userBalance.amount.amount - 5 : 5;
```
JustSamuel, line 192: is it 10 or 5 euros? in the mail you say 5

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,320 @@
...
+
+    const transfer = await new TransferService(this.manager).createTransfer(transferRequest);
+
+    // Create a new inactive administrative cost
+    const newInactiveAdministrativeCost: InactiveAdministrativeCost = Object.assign(new InactiveAdministrativeCost(), {
+      fromId: forId,
+      from: user,
+      amount: dinero(amount),
+      transfer: transfer,
+    });
```
JustSamuel, line 216: prefer `insert` or `save` as this now does not have any linting 

(I know, we do this wrong _everywhere_)

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,320 @@
...
+
+    transfer.inactiveAdministrativeCost = newInactiveAdministrativeCost;
+
+    await this.manager.getRepository(Transfer).save(transfer);
```
JustSamuel, line 220: see, here it is unable to determine that it is of type transfer

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,320 @@
...
+  public async getPaginatedInactiveAdministrativeCosts(params: InactiveAdministrativeCostFilterParameters = {},
+    pagination: PaginationParameters = {}) {
+    const { take, skip } = pagination;
+    const options = { ...InactiveAdministrativeCostService.getOptions(params), skip, take };
+
+    const inactiveAdministrativeCost = await this.manager.find(InactiveAdministrativeCost, { ...options, take });
+
+    const records = InactiveAdministrativeCostService.toArrayResponse(inactiveAdministrativeCost);
+
+    const count = await this.manager.count(InactiveAdministrativeCost, options);
```
JustSamuel, line 291: there is a find and count you can use

src/service/stripe-service.ts
```diff
@@ -57,7 +57,7 @@ export default class StripeService extends WithManager {
     this.stripe = new Stripe(process.env.STRIPE_PRIVATE_KEY, {
       apiVersion: STRIPE_API_VERSION,
     });
-    this.logger = getLogger('StripeController');
+    this.logger = log4js.getLogger('StripeController');
```
JustSamuel, line 60: why or is this a dep update?

src/service/transaction-service.ts
```diff
@@ -354,7 +354,9 @@ export default class TransactionService extends WithManager {
       ...update,
       version: update.version + 1,
       updatedAt: new Date(),
-    }) : Object.assign(new Transaction(), {})) as Transaction;
+    }) : Object.assign(new Transaction(), {
+      createdAt: req.createdAt ? new Date(req.createdAt) : new Date(),
+    })) as Transaction;
```
JustSamuel, line 359: why?

test/seed/ledger/inactive-administrative-cost-seeder.ts
```diff
@@ -0,0 +1,78 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ *  @license
+ */
+import WithManager from '../../../src/database/with-manager';
```
JustSamuel, line 20: space

test/unit/controller/inactive-administrative-cost-controller.ts
JustSamuel, line 1: will checker coverage later

JustSamuel (changes requested):

.env-example
```diff
@@ -18,7 +18,7 @@ JWT_KEY_PATH = config/jwt.key
 SKIP_SQLITE_DEFAULTS = false
 
 TYPEORM_CONNECTION = sqlite
-TYPEORM_DATABASE = local.sqlite
+TYPEORM_DATABASE = :memory:
```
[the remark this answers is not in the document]
SvenMokveldje: Small oepsie

src/controller/inactive-administrative-cost-controller.ts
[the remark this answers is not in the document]
SvenMokveldje: Agree that the name is quite long, but as discussed with Roy. If SudoSOS ever needs to implement another type of administrative costs, this makes it more clear.

src/controller/inactive-administrative-cost-controller.ts
```diff
@@ -0,0 +1,343 @@
...
+
+  /**
+   * GET /inactiveAdministrativeCosts
+   * @summary Returns all inactive administrative costs in the system.
+   * @operationId getAllInactiveAdministrativeCosts
+   * @tags inactiveAdministrativeCosts - Operations of the invoices controller
+   * @security JWT
+   * @param {integer} fromId.query - Filter on the id of the user
+   * @param {integer} inactiveAdministrativeCostId.query - Filter on the entity
+   * @param {boolean} notification.query - Boolean to check users for notification or for fine
```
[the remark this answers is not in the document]
SvenMokveldje: The param did not even make sense, thus has been removed

src/controller/inactive-administrative-cost-controller.ts
```diff
@@ -0,0 +1,343 @@
...
+      res.json(usersResponses);
+    } catch (error) {
+      this.logger.error('Could not check inactive users:', error);
+      res.status(500).json('Internal server error.');
+    }
+  }
+
+  /**
+   * POST /notify
+   * @summary Find all users who are eligible for notification or creation of inactive administrative cost
```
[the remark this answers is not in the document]
SvenMokveldje: Copy error

src/controller/inactive-administrative-cost-controller.ts
```diff
@@ -0,0 +1,343 @@
...
+   * @operationId notifyInactiveUsers
+   * @tags inactiveAdministrativeCosts - Operations of the invoices controller
```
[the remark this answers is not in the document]
SvenMokveldje: Copy error

src/controller/request/transaction-request.ts
```diff
@@ -34,13 +34,15 @@ import RevisionRequest from './revision-request';
  * @property {Array<SubTransactionRequest>} subTransactions.required - subtransactions
  * @property {RevisionRequest} pointOfSale.required - point of sale
  * @property {DineroObjectRequest} totalPriceInclVat.required - total price of the transaction
+ * @property {string} createdAt - Date on which the transfer should be created
```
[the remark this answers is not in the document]
SvenMokveldje: I wanted to make sure a user has old transactions to test on, but changed my test suite to just create old transaction and transfers while seeding.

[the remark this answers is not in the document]
JustSamuel: So why would you modify the request?

src/controller/inactive-administrative-cost-controller.ts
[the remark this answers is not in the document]
JustSamuel: I'll allow it.

src/service/inactive-administrative-cost-service.ts
SvenMokveldje: creditTransfer but I have changed this.

src/service/inactive-administrative-cost-service.ts
SvenMokveldje: Changed it a bit and it works but not certain what you mean.

SvenMokveldje (commented):
@JustSamuel these are my remarks back will push now.

.env-example
```diff
@@ -18,7 +18,7 @@ JWT_KEY_PATH = config/jwt.key
 SKIP_SQLITE_DEFAULTS = false
 
 TYPEORM_CONNECTION = sqlite
-TYPEORM_DATABASE = :memory:
+TYPEORM_DATABASE = local.sqlite
```
JustSamuel, line 21: >:(

src/service/inactive-administrative-cost-service.ts
SvenMokveldje: This is the same way as in the other services which have been updated with extends WithManager

Yoronex:
Please **rebase** your branch on the `develop` branch. Your PR includes for example all the changes I made in the dependency update, which should not be part of this PR (as it is already present in develop). Not sure why this goes wrong, but please try to rebase instead of merging develop into your own branch when it is outdated.

SvenMokveldje:
@JustSamuel @Yoronex, I think I have implemented your changes. Please check :)

src/service/inactive-administrative-cost-service.ts
JustSamuel: No, but I do would like a comment above it since the code is heard to read.

src/migrations/1729266426571-user-administrative-cost.ts
JustSamuel, line 1: looks fine, but will need some testing.

SvenMokveldje: Yeah, I think the post migration test will fail, but I do not understand it 100%. I do think the rest should now be correct.

JustSamuel (commented):
small request

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,327 @@
...
+        relations: { inactiveAdministrativeCost: true },
+      }));
+
+      // Checks whether a user has transactions or transfers and reduces over them all to find the youngest.
+      const lastTransfer = userTransfers.length == 0 ? null : userTransfers
+        .reduce((prev, curr) => (((prev.createdAt < curr.createdAt) && curr.inactiveAdministrativeCost == null) ? curr : prev));
+      const userTransactions = ((await Transaction.find({ relations: ['from'] }))
+        .filter((t) => t.from.id === user.id));
+      const lastTransaction = userTransactions.length == 0  ? null : userTransactions
+        .reduce((prev, curr) => (prev.createdAt < curr.createdAt ? curr : prev));
```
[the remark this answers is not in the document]
JustSamuel: example [here](https://github.com/GEWIS/sudosos-backend/blob/dabd0e5ae0edfa84e386e7cc9decbaa5a13388e3/src/service/invoice-service.ts#L532)

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,327 @@
...
+  public async createInactiveAdministrativeCost(inactiveAdministrativeCostRequest: CreateInactiveAdministrativeCostRequest)
+    : Promise<InactiveAdministrativeCost> {
+    const { forId } = inactiveAdministrativeCostRequest;
+
+    // Calculate reduction amount
+    const user = await this.manager.findOne(User, { where: { id: forId } });
+    const userBalance = await new BalanceService(this.manager).getBalance(forId);
+
+    const monetaryAmount = (userBalance.amount.amount < 10) ? userBalance.amount.amount - 10 : 10;
+
```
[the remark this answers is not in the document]
JustSamuel: example [here](https://github.com/GEWIS/sudosos-backend/blob/dabd0e5ae0edfa84e386e7cc9decbaa5a13388e3/src/service/write-off-service.ts#L120)

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,327 @@
...
+      toId: inactiveAdministrativeCost.fromId,
+    };
+
+    // Save new transfer and delete the administrative cost
+    await new TransferService(this.manager).postTransfer(undoTransfer).then(async (response) => {
+      const transfer = await this.manager.findOne(Transfer, { where: { id: response.id } });
+      if (!transfer) throw new Error('Transfer not found during deletion of inactive administrative cost, aborting');
+      inactiveAdministrativeCost.creditTransfer = transfer;
+    });
+    await this.manager.save(InactiveAdministrativeCost, inactiveAdministrativeCost);
```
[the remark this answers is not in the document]
SvenMokveldje: No because the response is of type transferresponse not transfer

JustSamuel:
@SvenMokveldje small thing I thought about, if we want to deploy this we also need to think about the impact. This is something the BACPM will need to add to her bookkeeping, and therefore should already come with the pdf functionality implemented. 

We could make it a different PR, but before deploying this to main it should be possible to get a pdf overview of the handed out administration costs.

Also maybe an interface on the frontend side of things? What do you think.

SvenMokveldje:
@JustSamuel I can still look into the PDF part as part of this PR as I agree that this should be looked into. It is however a very small things that only happens once a year (at least my idea was to always check it on the first of July or maybe once a quartile). Next to that from what I see, people that do not use it will mostly likely be disabled first by having too many fines.

So, I think in the frontend you could add this to one of the already existing overviews?

JustSamuel:
quick question @SvenMokveldje, the whole process now works with that the BACPM remembers to do this right? As in everything is done manually?

Maybe we should consider a reminder mail for the BACPM, at the first of july, weekly, until its done?

src/controller/response/inactive-administrative-cost-response.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * @typedef {object} UserToInactiveAdministrativeCostResponse
+ * @property {integer} id.required - User ID
+ */
+export interface UserToInactiveAdministrativeCostResponse {
+  userId: number;
+}
+
+/**
+ * @typedef {object} PaginatedInactiveAdministrativeCost
```
JustSamuel: name should be PaginatedInactiveAdministrativeCostResponse

src/entity/user/user.ts
```diff
@@ -70,6 +70,13 @@ export const NotifyDebtUserTypes: UserType[] = [
   UserType.LOCAL_ADMIN, UserType.LOCAL_USER, UserType.MEMBER,
 ];
 
+/**
+ * All users that have made inactive administrative costs.
+ */
+export const EligibleInactiveUsers: UserType[] = [
+  UserType.LOCAL_ADMIN, UserType.LOCAL_USER, UserType.MEMBER,
```
JustSamuel: LOCAL_ADMIN? not to sure.

src/service/inactive-administrative-cost-service.ts
```diff
@@ -0,0 +1,343 @@
...
+      .where('inactiveAdministrativeCost.Id is NULL')
+      .andWhere('transfer.fromId = :userId', { userId })
+      .getOne();
+  }
+
+  private async lastTransactionQuery(userId: number): Promise<Transaction> {
+    return Transaction.getRepository()
+      .createQueryBuilder('transaction')
+      .select('MAX(createdAt) as createdAt')
+      .where(`transaction.from.id = ${userId}`)
```
JustSamuel: I don't think this is how that works?

the `transaction.from.id` looks VERY suspicious.

JustSamuel (changes requested):
Some feedback.

ALso make sure that you add the correct permissions to the `default roles` in the RBAC.

src/controller/inactive-administrative-cost-controller.ts
```diff
@@ -0,0 +1,344 @@
...
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ *  @license
+ */
+
+import BaseController, { BaseControllerOptions } from './base-controller';
```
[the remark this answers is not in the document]
SvenMokveldje: Added

src/service/inactive-administrative-cost-service.ts
SvenMokveldje: Yes should be transaction.fromId (column in the database)

src/database/database.ts
```diff
@@ -116,6 +120,7 @@ const options: DataSourceOptions = {
   migrations: [
     InitialSQLMigration1743601882766,
     QrAuthenticator1743601882766,
+    UserAdministrativeCost1760553313387,
```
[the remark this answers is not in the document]
SvenMokveldje: Do not know exactly what date /which number to change then

JustSamuel (changes requested):
Still missing the default permissions in the RBAC

Issue #335, opened 2024-09-26
[Feature] Waiving fines should be a custom amount, instead of the total amount of fines


Yoronex:
### What would you like?

See title

### Why is this needed?

The treasurer's goal is to make people pay for their beer. Fines are a means to an end. However, sometimes people collect such high fines that they are not willing to pay anymore. Then, the treasurer makes a deal to waive a part of the total fine. SudoSOS should implement this, i.e. allow the waived amount to be between 0 and the total value of fines.

### Other information

Currently a blocking issue for the treasurer to get rid of debts, so high priority.

Issue #334, opened 2024-09-26
[Feature] Balance responses should extend `BaseUserResponse`


Yoronex:
### What would you like?

The `BalanceResponse` should extend the `BaseUserResponse`.

### Why is this needed?

If we include such information, it is much easier to see to which user a fetched balance actually belongs. This is especially relevant around handing out fines, so the frontend can easily show each user's balance and total fine.

Yoronex:
These issues can probably implemented in the same PR:
https://github.com/GEWIS/sudosos-backend/issues/346
https://github.com/GEWIS/sudosos-backend/issues/352

PR #344, opened 2024-10-06
feat: partially waive user's fines

Yoronex:
# Description

After merging the PR, a `UserFineGroup`'s can be partially waived, instead of only all of them. The following decision decisions/restrictions are made:
1. The workflow of waiving all user's fines has not changed. When all fines of a user are waived, the user will no longer have any outstanding fines.
2. When waiving fines partially, the user will no longer have any outstanding fines if they go out of debt after the waiving.
3. When a user gets its fines partially waived and remains in debt, the user will still have outstanding fines. If they get a new fine, this new fine will be added to the total outstanding amount (the same `UserFineGroup`).
4. When a user's fines of the same `UserFineGroup` are waived _again_, this new amount to waive will override the old amount. In other words, the old `waiveFineTransfer` is deleted and a new transfer is created.
5. For backwards compatibility with the current SudoSOS API spec, the waive amount is optional. When no amount is provided, the total fine amount will be waived. This behaviour should be seen as deprecated.

Note that the amount of fines responded by the `BalanceController` is still the total amount of fines and does not include any waived amounts. Having these numbers separately is required to support point 4 (so you know the total amount of fines and how much is already waived). This waived amount should be implemented as a seperate field in `BalanceResponse` and can probably be best done together with the rest of https://github.com/GEWIS/sudosos-backend/issues/334.

https://github.com/GEWIS/sudosos-backend/issues/335

## Types of changes

### DIFF
--- a/src/controller/user-controller.ts
+++ b/src/controller/user-controller.ts
@@ -70,6 +70,7 @@ import ReportService, { BuyerReportService, SalesReportService } from '../servic
 import { ReturnFileType, UserReportParametersType } from 'pdf-generator-client';
 import { reportPDFhelper } from '../helpers/express-pdf';
 import { PdfError } from '../errors';
+import { WaiveFinesRequest } from './request/debtor-request';
 
 export default class UserController extends BaseController {
   private logger: Logger = log4js.getLogger('UserController');
@@ -330,6 +331,7 @@ export default class UserController extends BaseController {
         POST: {
           policy: async (req) => this.roleManager.can(req.token.roles, 'delete', 'all', 'Fine', ['*']),
           handler: this.waiveUserFines.bind(this),
+          body: { modelName: 'WaiveFinesRequest' },
         },
       },
     };
@@ -1602,6 +1604,7 @@ export default class UserController extends BaseController {
    * @summary Waive all given user's fines
    * @tags users - Operations of user controller
    * @param {integer} id.path.required - The id of the user
+   * @param {WaiveFinesRequest} request.body.required
    * @operationId waiveUserFines
    * @security JWT
    * @return 204 - Success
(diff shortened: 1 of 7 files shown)


SvenMokveldje (commented):
No comments but will leave the approval to @JustSamuel

src/controller/user-controller.ts
```diff
@@ -1610,14 +1615,13 @@ export default class UserController extends BaseController {
    */
   public async waiveUserFines(req: RequestWithToken, res: Response): Promise<void> {
     const { id: rawId } = req.params;
-    this.logger.trace('Waive fines of user', rawId, 'by', req.token.user);
```
JustSamuel, line 1613: This removes the ability to waive all fines in a single call right?

src/controller/user-controller.ts
```diff
@@ -1627,7 +1631,19 @@ export default class UserController extends BaseController {
         return;
       }
 
-      await DebtorService.waiveFines(id);
+      const totalAmountOfFines = user.currentFines!.fines.reduce((total, f) => total.add(f.amount), Dinero());
+      // Backwards compatibility with old version, where you could only waive all user's fines
```
JustSamuel, line 1635: Related to my previous remark, why not split it into two routes?

JustSamuel (approved):

src/controller/user-controller.ts
Yoronex: No, this is still possible.
- As part of the transition to this new format, not sending any body to the endpoint will behave the same as it does now, namely waive all fines.
- When you give the total amount of fines as amount to waive, all fines will be waived.

src/controller/user-controller.ts
Yoronex: Because you can still waive all fines with this endpoint, but you have to give the total amount of fines manually. Making two endpoints will in my opinion just be confusing, as both can do the same thing. When do you use which endpoint?

Issue #347, opened 2024-10-08
[Chore] All Event-related modules/classes should be marked as deprecated


Yoronex:
### What would you like?

See title

### Why is this needed?

At the ABC SudoSOS meeting, we have decided that events are out of scope for SudoSOS.

### How could it be implemented?

To preserve the existing (functional) code for a potential future project, the code should be marked as deprecated with a "delete-from" starting two years for now. After this date, all these modules can be deleted permanently (because the code probably won't be used anymore).

PR #356, opened 2024-10-17
chore: deprecate all Event-related code

Yoronex:
# Description

I have marked all event-related code as deprecated to the best of my abilities. I also added a comment in the deprecation that these deprecations can be removed from November 1st 2026 onwards (two years from now). If we have never moved this code to another project before that time, I assume it will never happen.

Fixes https://github.com/GEWIS/sudosos-backend/issues/347.

## Types of changes


JustSamuel (approved):
o7

PR #381, opened 2024-10-28
fix(docker): reduce size of image by about half

SuperVK:
During the Ubuntu Summit I was lead down a rabbit hole of inspecting docker images. I noticed that the sudosos-backend could use some improvement.

# Description

I noticed that the npm ci still installed dev dependencies in the runtime container, so that was fixed. Furthermore when installing npm packages, it also adds to the cache. Meaning it used to have 2x the node_modules installed in the image. By installing in the build container and copying to the runtime container, we circumvent this issue.

image
The build from 6 days ago is the current develop docker image, the docker image from 9 minutes ago is this the one that has the improvement. 

Here you can see the cache being added which is about the same size as the node_modules themself:
image

My improved image is currently running on develop (28-10-2024 10:53), it seems to work, only the pdf is broken, but pretty sure that has always been broken on develop?

## Types of changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN npm install
 COPY ./ ./
 RUN npm run build \
  && npm run swagger
+RUN npm ci --production
 
 # The target image that will be run
 FROM node:18-alpine as target
@@ -13,8 +14,7 @@ FROM node:18-alpine as target
 RUN apk add openssl
 
 WORKDIR /app
-COPY ./package.json ./package-lock.json ./
-RUN npm ci
+COPY --from=build --chown=node /app/node_modules /app/node_modules
 RUN npm install pm2 pm2-graceful-intercom -g
 RUN npm install -g typeorm
 ARG TYPEORM_USERNAME


SuperVK:
It would be ideal if some build tool could minify the dependencies and only include what is needed, like vite does for the frontend. But that's not super needed, and I think requires quite a lot more changes.

JustSamuel:
@SuperVK we could switch to vite?

JustSamuel (approved):

Yoronex:
I don't really follow why this reduces the size of the image, probably because you do `npm ci --production`. Why can't we just do run that command in the target image? Why do we have to run this command in the build image and then copy it to the target? This does not make sense to me, and for performance reasons, I would argue you would do the former, as that allows both `npm ci` commands to be run in parallel.

JustSamuel:
> ally follow why this reduces the size of the image, probably because you do `npm ci --production`. Why can't we just do run that command in the target image? Why do we have to run this command in the build image and then copy it to the target? This does not make sense to me, and for performance reasons, I would argue you would do the former, as that allows both `npm ci` commands to be run in parallel.

Running `npm ci` (even `--production`) might install build tools. So no, copy is the way to go here.

PR #419, opened 2025-01-09
feat(database): allow enforcing of SSL for database connection

tomudding:
# Description

Several ways devised to implement this easily. Most logical is to enforce this feature only when actually running in production, but this does not work well when using CI for testing. So a `TYPEORM_SSL_ENABLED` environment variable provides more configuration flexibility (so you can also test this locally).

Originally, I wanted to do it this way:

```typescript
const sslConfig: any = process.env.TYPEORM_SSL === 'production' ? {
  ssl: {
    ca: fs.readFileSync('/etc/ssl/certs/ca-certificates.crt')
  }
} : {};
```

I thought about the potential complications of having `ssl: {},` but I do not think it will cause any issues. I did try an alternative (smart) solution, but the type checker does not agree :'( Even when casting `as MysqlConnectionOptions | PostgresConnectionOptions` it does not work because it starts complaining about the fact that the `ssl` property is `readonly`. So it must be inlined.

Originally, the path to the CA certificates file was hardcoded. However, this does not allow for flexibility on non-standard platforms that store it in a different location or specific requirements from the environment. As such, the `TYPEORM_SSL_CACERTS` has been introduced to handle that.

See the e-mail from the CBC.

## Types of changes


tomudding:
@JustSamuel, I want specific feedback on whether to make the path to the CA certificates (also) an environment variable.

Hardcoding is fine for the production images, as they always target a Linux distro. However, if someone decides to work on SudoSOS locally using an exotic distro that has things differently, it breaks. In that case, moving the path to an environment variable would make sense. For ease of use, I think it would make sense to make the `TYPEORM_SSL` (introduced in this PR), not a boolean but a path. Then, if it is set SSL is enabled with that path, if it is not set SSL is not enabled.

JustSamuel:
> Linux distro. However, if someone decides to work on SudoSOS locally using an exotic distro that has things differently, it breaks. In that case, moving the path to an environment variable would make sense. For ease of use, I think it would make sense to make the `TYPEORM_SSL` (introduced in this PR), not a boolean but a path. Then, if it is set SSL is enabled with that path, if it is not set SSL is not enabled.

I am personally more in favor of making sure there is a nice sensible default / fallback value. How about having a `TYPEORM_SSL_ENABLED` for toggling SSL and using `TYPEORM_SSL_CERT` if defined, otherwise use the default? Being verbose in this way is not an issue imho.

JustSamuel (approved):

PR #423, opened 2025-01-12
fix: rbac-perm-deletion-wrong-type

SvenMokveldje:
The controller had the wrong type as input, namely for entity, relation and rule the input was integer but this should've been a string.


RubenLWF (approved):
🚀

Issue #422, opened 2025-01-10
NFC user search when logged into the POS


noelvandevrie:
### What would you like?

To be able to search users by NFC code when te POS is logged in.

### Why is this needed?

It will add additional ways to pay for things for each other outside of borrels, and during borrels the BAC could ask borrelaars for their linked NFC card and find users this way.

### How could it be implemented?

NFC Searching should be an endpoint, personally I don't know where exactly this should take place.

JustSamuel:
I propose `GET users/nfc/{id}` and return a 404 or a 200

Yoronex:
Implemented in https://github.com/GEWIS/sudosos-backend/pull/424.

PR #424, opened 2025-01-13
feat(users): add endpoint to lookup user from nfc code

JustSamuel:
# Description
Adds an endpoint to the user-controller to lookup a user by nfc code. Intented implementation will be for the POS to allow charging a user by simply scanning their nfc code.

#422 

## Types of changes

### DIFF
--- a/src/controller/user-controller.ts
+++ b/src/controller/user-controller.ts
@@ -133,6 +133,14 @@ export default class UserController extends BaseController {
           restrictions: { acceptedTOS: false },
         },
       },
+      '/nfc/:nfcCode': {
+        GET: {
+          policy: async (req) => this.roleManager.can(
+            req.token.roles, 'get', 'all', 'User', ['*'],
+          ),
+          handler: this.findUserNfc.bind(this),
+        },
+      },
       '/:id(\\d+)/authenticator/pin': {
         PUT: {
           body: { modelName: 'UpdatePinRequest' },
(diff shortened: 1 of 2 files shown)


src/controller/user-controller.ts
```diff
@@ -133,6 +133,14 @@ export default class UserController extends BaseController {
           restrictions: { acceptedTOS: false },
         },
       },
+      '/nfc/:id': {
```
Yoronex, line 136: Calling this parameter `id` was pretty confusing to me at first. Maybe rename it to something else like `nfcCode` or something, so it is also not confusing in generated clients.

src/controller/user-controller.ts
```diff
@@ -876,6 +884,42 @@ export default class UserController extends BaseController {
...
+   * @return {UserResponse} 200 - The requested user
+   * @return {string} 404 - The user with the given nfc code does not exist
+   */
+  public async findUserNfc(req: RequestWithToken, res: Response): Promise<void> {
+    const parameters = req.params;
+    this.logger.trace('Find user nfc', parameters, 'by user', req.token.user);
+
+    try {
+      const nfcCode = String(parameters.id);
+      const nfc = await NfcAuthenticator.findOne({ where: { nfcCode } });
```
Yoronex, line 903: Can't you just use `{ where: { nfcCode }, relations: { user: true } }` such that the user is automatically returned? That makes the condition on line 911 obsolete (which also does not really make sense, as all NFC authenticators must have a corresponding user, but I see why you had to do it).

Yoronex (approved):

src/controller/user-controller.ts
JustSamuel: Kinda, the UserService has some tech debt, so this will mean that the response will always be a UserResponse and not a GEWIS UserResponse. Fixing this is outside the scope of this PR, and is therefore a trade-off.

How important is that the GEWIS id is also returned?

PR #458, opened 2025-02-12
fix: contain delete propagated to wrong pos

JustSamuel:
kutzooi


tomudding (approved):
LGTM (and same changes as for GH-451)

PR #464, opened 2025-02-12
Allowing `getOptions` to return deleted products

JustSamuel:
ffs.


src/service/transaction-service.ts
```diff
@@ -138,7 +138,8 @@ export default class TransactionService extends WithManager {
   public async getTotalCost(rows: SubTransactionRowRequest[]): Promise<Dinero.Dinero> {
     // get costs of individual rows
     const rowCosts = await Promise.all(rows.map(async (row) => {
-      const options =  await ProductService.getOptions({ productRevision: row.product.revision, productId: row.product.id });
+      const options =  await ProductService.getOptions({ productRevision: row.product.revision,
+        productId: row.product.id, allowDeleted: true });
```
tomudding: For readability, I would split this for each selector.

tomudding (approved):

Issue #466, opened 2025-02-12
Do not invalidate balance cache for user when creating transaction


tomudding:
### Current behavior

Currently, the balance cache is cleared for a user when a transaction for them is created/updated/deleted. However, the balance cache should not be invalidated when _creating_ a transaction:

https://github.com/GEWIS/sudosos-backend/blob/8ffb1ee0cf031c2dc0fa60d1368d346a265d3f53/src/service/transaction-service.ts#L664-L676

This is an action forward; the new balance is just the sum of what is in the balance cache + all new transactions for that user. So there is no need to remove it. Furthermore, it is bad for performance, as from that point forward until the next balance cache calculation (cronjob), the balance has to be calculated on the fly.

This will also slow down the new debtor overview, as many people will have their balance cache entry removed after a social drink.

### Desired behavior

The balance cache is not removed when `creating` a transaction, only when _updating_ or _deleting_ a transaction.

### Steps to reproduce

1. Calculate balance cache
2. Create a transaction for a user
3. Observe balance cache entry removed for the user

PR #467, opened 2025-02-12
fix: balance cache invalidation on transaction creation

tomudding:
# Description

Currently, balance cache is cleared when a transaction is created/updated/ deleted for a user. However, balance cache should not be invalidated when creating a transaction.

Creating a transaction is an action forward; new balance is just the sum of balance cache + all new transactions for that user. Thus, no need to remove it. Invalidating it is also bad for performance, as balance has to be calculated on the fly until next balance cache calculation (cronjob).

This removes balance cache invalidation for transaction creations.

---

Here confirmation of the old behaviour:
image
After creating a transaction:
image

GH-466

## Types of changes

### DIFF
--- a/src/service/transaction-service.ts
+++ b/src/service/transaction-service.ts
@@ -667,12 +667,8 @@ export default class TransactionService extends WithManager {
 
     await transaction.save();
 
-    // save the transaction and invalidate user balance cache
-    const savedTransaction = await this.asTransactionResponse(transaction);
-    await TransactionService.invalidateBalanceCache(savedTransaction);
-
     // save transaction and return response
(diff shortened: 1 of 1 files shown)


JustSamuel (approved):
I am still kinda worried why this was ever included? Do we even actually update the balance cache after this?

tomudding:
> I am still kinda worried why this was ever included? Do we even actually update the balance cache after this?

I cannot answer the first question (see introduction in 3b24477e9f6b62a7c763663fe241d64d2d395b7d), but it is updated/repopulated as expected at 01:14:
image

PR #469, opened 2025-02-14
Feature: maintenance mode websocket

RubenLWF:
# Description

This PR creates a websocket server that we will send a message over once the `PUT /server-settings/maintenance-mode` endpoint is called. This message can be used by the POS to show the maintenance screen as soon as maintenance mode is enabled without having to be refreshed.

## Types of changes


JustSamuel:
Nice to see you are working on the backend! :D

Small issue / food for thought about this new feature though, in production we run 4 processes for the main endpoints and one for cron jobs. The API calls are distributed evenly over these 4 processes, which is nice for high load situtions.

The issue, as you might guess, is that the process that receives the server-setting update might not be the same as the one storing the clients (in the end, you are only connected to a single process with the websocket).

Possible solutions include stuff like redis or storing jobs in the db, but I think this solution works best:
https://socket.io/docs/v4/pm2/



Also, second remark. I think it would be a bit nicer if we create some pub-sub type structure? (and possible a authentication layer or handshake) In the end it would be nice to use this to get live transaction history on the POS without having to pol continuously. To this end I suggest a structure where you can connect to specific topic (transactions, topups, etc). However it should not be so complex that we are reworking our CRUD endpoints into pubsub topics, but I think a set of predefined topics could get is quite a distance.

src/service/websocket-service.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @module websocket
+ */
+
+export default class WebSocketService {
+  private static readonly httpServer = http.createServer(express());
+
+  private static webSocketServer = new WebSocket.Server({ server: this.httpServer });
+    
+  public static initiateWebSocket(): void {
+    const logger: Logger = log4js.getLogger('LDAP');
```
JustSamuel, line 39: LDAP

RubenLWF:
This seems to work, any comments before I start working on tests?

src/service/websocket-service.ts
```diff
@@ -0,0 +1,64 @@
...
+      });
+    }
+
+    this.IO.on('connection', (socket) => {
+      logger.info(`connect ${socket.id}`);
+    });
+  }
+
+  public static sendMaintenanceMode(enabled: boolean): void {
+    const logger: Logger = log4js.getLogger('WebSocket');
```
JustSamuel, line 59: can be static, not?

JustSamuel (changes requested):
Would it be possible to add e2e test cases?

JustSamuel (approved):

PR #495, opened 2025-03-13
feat: total-balances-sudosos

SvenMokveldje:
Added a balance service function to get a added up value of positive and negative balances. you can also decide to do this for all users or only for users that can top up (member and local)

# Description
This should decrease the time it takes to calculate this as in the frontend this now takes +- 14 seconds.

## Types of changes

### DIFF
--- a/src/controller/balance-controller.ts
+++ b/src/controller/balance-controller.ts
@@ -71,6 +71,12 @@ export default class BalanceController extends BaseController {
           handler: this.getBalance.bind(this),
         },
       },
+      '/summary': {
+        GET: {
+          policy: async (req) => this.roleManager.can(req.token.roles, 'get', 'all', 'Balance', ['*']),
+          handler: this.calculateTotalBalances.bind(this),
+        },
+      },
     };
   }
 
@@ -176,6 +182,34 @@ export default class BalanceController extends BaseController {
     } catch (error) {
       const id = req?.params?.id ?? req.token.user.id;
       this.logger.error(`Could not get balance of user with id ${id}`, error);
+      res.status(500).json('Internal server error.');
+    }
+  }
+
+  /**
+   * GET /balances/summary
+   * @summary Get the calculated total balances in SudoSOS
+   * @operationId calculateTotalBalances
+   * @tags balance - Operations of balance controller
+   * @security JWT
+   * @param {string} date.query.required - The date for which to calculate the balance.
+   * @return {TotalBalanceResponse} 200 - The requested user's balance
+   * @return {string} 400 - Validation error
+   * @return {string} 404 - Could not calculate error
+   * @return {string} 500 - Internal server error
+   */
+  private async calculateTotalBalances(req: RequestWithToken, res: Response): Promise<void> {
+    try {
+      const date = asDate(req.query.date);
+
+      try {
+        const balances = await new BalanceService().calculateTotalBalances(date);
+        res.json(balances);
+      } catch (error) {
+        res.status(404).json('Could not calculate the total balances');
+      }
+    } catch (error) {
+      this.logger.error('Could not calculate the total balances', error);
       res.status(500).json('Internal server error.');
     }
   }
--- a/src/controller/response/balance-response.ts
+++ b/src/controller/response/balance-response.ts
@@ -63,6 +63,32 @@ export default interface BalanceResponse extends BaseUserResponse {
 }
 
 /**
+ * @typedef {object} UserTypeTotalBalanceResponse
+ * @property {string} userType - The user type
+ * @property {DineroObjectResponse} totalPositive - The total amount of positive balance for this user type
+ * @property {DineroObjectResponse} totalNegative - The total amount of negative balance for this uer type
+ */
+export interface UserTypeTotalBalanceResponse {
+  userType: UserType,
+  totalPositive: DineroObjectResponse,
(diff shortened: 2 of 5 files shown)


src/controller/balance-controller.ts
```diff
@@ -71,6 +71,12 @@ export default class BalanceController extends BaseController {
           handler: this.getBalance.bind(this),
         },
       },
+      '/calculate': {
```
JustSamuel, line 74: not really `RESTful` , how about `summary` or `total`?

src/controller/balance-controller.ts
```diff
@@ -179,4 +185,40 @@ export default class BalanceController extends BaseController {
...
+   * @param {date} date.path.required - The date for which to calculate the balance.
+   * @return {TotalBalanceResponse} 200 - The requested user's balance
+   * @return {string} 400 - Validation error
+   * @return {string} 404 - Could not calculate error
+   * @return {string} 500 - Internal server error
+   */
+  private async calculateTotalBalances(req: RequestWithToken, res: Response): Promise<void> {
+    try {
+      const allUsers = Boolean(req.query.allUsers);
+      const dateString = req.query.date as string;
```
JustSamuel, line 205: `asDate`?

src/controller/response/balance-response.ts
```diff
@@ -62,6 +62,18 @@ export default interface BalanceResponse extends BaseUserResponse {
   lastTransferId: number;
 }
 
+/**
+ * @typedef TotalBalanceResponse
```
JustSamuel, line 66: missing `{object}`

src/controller/balance-controller.ts
```diff
@@ -179,4 +185,40 @@ export default class BalanceController extends BaseController {
...
+
+      const date = new Date(dateString);
+
+      if (isNaN(date.getTime())) {
+        res.status(400).json('Invalid date format.');
```
JustSamuel, line 210: missing return

src/controller/response/balance-response.ts
```diff
@@ -62,6 +62,18 @@ export default interface BalanceResponse extends BaseUserResponse {
...
+ * @property {string} data.required - Date at which this total balance was calculated
+ * @property {number} totalPositive.required - The total amount of positive balance in SudoSOS
```
JustSamuel, line 68: why not a dinero object? seems more consistent with rest of monetairy responses.

src/service/balance-service.ts
```diff
@@ -380,4 +381,30 @@ export default class BalanceService extends WithManager {
...
     return (await this.getBalances({ ids: [id], allowDeleted: true, date })).records[0];
   }
+
+  /**
+   * Get the total balance of SudoSOS
+   * @param allUsers Whether or not you want to get the balance for all users or only the users that can top up.
+   * @param date Date to check the balances for
+   */
+  public async calculateTotalBalances(allUsers: boolean, date: Date): Promise<TotalBalanceResponse> {
+    const userIds = allUsers ? (await User.find()).map((user) => user.id)
```
JustSamuel, line 391: I think a single `await User.find()` would be nicer, and instead making the where a conditional, as this is quite hard  to read.

```ts
const where = allUsers ? {} : { type: In([UserType.LOCAL_USER, UserType.MEMBER]) }
```

src/service/balance-service.ts
```diff
@@ -380,4 +381,30 @@ export default class BalanceService extends WithManager {
...
+      .records.map((balance) => balance.amount.amount);
+
+    const totalPositive = positiveBalances.reduce((sum, num) => sum + num, 0);
+    const totalNegative = negativeBalances.reduce((sum, num) => sum + num, 0);
+
+    return {
+      date: date.toISOString(),
+      totalPositive:  totalPositive,
+      totalNegative: totalNegative,
+    };
```
JustSamuel, line 408: So possibly return as a dineroObjectResponse instead

JustSamuel (changes requested):
Small changes

SvenMokveldje:
Verbetering: altijd totaal laten zien en dan uitsplitsen op usertype.

Response {
total {pos, neg}
usertype {
member {pos, neg}
}

src/controller/response/balance-response.ts
```diff
@@ -62,6 +62,32 @@ export default interface BalanceResponse extends BaseUserResponse {
...
+  totalPositive: DineroObjectResponse,
+  totalNegative: DineroObjectResponse,
+}
+
+/**
+ * @typedef {object} TotalBalanceResponse
+ * @property {string} date - Date at which this total balance was calculated
+ * @property {number} totalPositive - The total amount of positive balance in SudoSOS
+ * @property {number} totalNegative - The total amount of negative balance in SudoSOS
+ * @property {UserTypeTotalBalanceResponse} userTypeBalances - The total balances for the different user types
```
JustSamuel, line 82: all missing a `.required`

```suggestion
 * @property {string} date.required - Date at which this total balance was calculated
 * @property {number} totalPositive.required - The total amount of positive balance in SudoSOS
 * @property {number} totalNegative.required - The total amount of negative balance in SudoSOS
 * @property {UserTypeTotalBalanceResponse} userTypeBalances.required - The total balances for the different user types
```

JustSamuel (approved):
small suggestions

PR #510, opened 2025-04-05
fix: balance-controller

SvenMokveldje:


src/controller/balance-controller.ts
```diff
@@ -195,19 +195,15 @@ export default class BalanceController extends BaseController {
    * @param {string} date.query.required - The date for which to calculate the balance.
    * @return {TotalBalanceResponse} 200 - The requested user's balance
    * @return {string} 400 - Validation error
-   * @return {string} 404 - Could not calculate error
+   * @return {string} 404 - Not found error
```
JustSamuel, line 198: This will never throw a 404 right?

JustSamuel (changes requested):

JustSamuel (approved):

PR #537, opened 2025-05-17
Feature/get role users

SvenMokveldje:
# Description
Created an endpoint to get all users related to a certain role


SvenMokveldje:
This is not yet complete, I think the functionality is correct but all tests are still missing. As I'm uncertain how to continue. Currently no users are seeded to assignments hence now I just assign 1 user to  a role. I would like to check whether I should create an extra seed function or add a User parameter to the RBAC seeder.

src/controller/rbac-controller.ts
```diff
@@ -78,6 +78,12 @@ export default class RbacController extends BaseController {
           handler: this.deleteRole.bind(this),
         },
       },
+      '/roles/:id(\\d+)/users': {
+        GET: {
+          policy: async () => true,
```
JustSamuel, line 83: Should be get all users I suppose?

src/controller/rbac-controller.ts
```diff
@@ -148,6 +154,36 @@ export default class RbacController extends BaseController {
...
+  public async getRoleUsers(req: RequestWithToken, res: Response): Promise<void> {
+    const { id } = req.params;
+    this.logger.trace('Get all users of role', id, 'by user', req.token.user);
+
+    try {
+      const roleId = Number(id);
+
+      const [[role]] = await RBACService.getRoles({ roleId });
+      if (!role) {
+        res.status(404).json('Role not found.');
```
JustSamuel, line 176: missing return

src/service/rbac-service.ts
```diff
@@ -168,6 +171,19 @@ export default class RBACService {
...
+   * Gets all users which are assigned to a certain role
+   * @param roleId
+   * @param take
+   * @param skip
+   */
+  public static async getRoleUsers(roleId: number, { take, skip }: PaginationParameters = {}): Promise<PaginatedUserResponse> {
+    const assignedRoles = await AssignedRole.find({ where: { roleId: roleId } });
+
+    const userIds = assignedRoles.map((role) => role.userId);
+    return UserService.getUsers({ id: userIds }, { take, skip });
```
JustSamuel, line 184: This does kinda remove all the logic for pagination? 

As now assignedRoles is the COMPLETE data, and only then we paginate. Pagination should be on query level if possible.

JustSamuel:
I would add optinal users to the RBAC seeding which wel get custom roles.

Then, in this test suite, you could create "assignedUsers = <rbac seeded with users>" or something like that.

src/service/rbac-service.ts
SvenMokveldje: Agree

SvenMokveldje:
@JustSamuel For some reason the user controller get users with a query fails. Do not know how that has happened as I did not touch it in any way, next to that running it and testing it manually seems to work and not give a bad request.

Maybe you know more?

src/controller/rbac-controller.ts
JustSamuel: Still needs to be fixed.

JustSamuel (changes requested):
Still need to fix the policy

src/controller/rbac-controller.ts
JustSamuel: 💀

JustSamuel (changes requested):
Please change the policy

SvenMokveldje:
I have changed the policy to the same as the get users policy.

src/controller/rbac-controller.ts
```diff
@@ -78,6 +78,14 @@ export default class RbacController extends BaseController {
           handler: this.deleteRole.bind(this),
         },
       },
+      '/roles/:id(\\d+)/all-users': {
```
JustSamuel, line 81:
```suggestion
      '/roles/:id(\\d+)/users': {
```

JustSamuel (changes requested):

src/controller/rbac-controller.ts
```diff
@@ -148,6 +156,37 @@ export default class RbacController extends BaseController {
     }
   }
 
+  /**
+   * GET /rbac/roles/{id}/all-users
```
JustSamuel, line 160:
```suggestion
   * GET /rbac/roles/{id}/users
```

JustSamuel (approved):

test/unit/controller/rbac-controller.ts
```diff
@@ -214,6 +219,35 @@ describe('RbacController', async (): Promise<void> => {
     });
   });
 
+  describe('GET /rbac/roles/{id}/all-users', () => {
```
JustSamuel, line 222:
```suggestion
  describe('GET /rbac/roles/{id}/users', () => {
```

test/unit/controller/rbac-controller.ts
```diff
@@ -214,6 +219,35 @@ describe('RbacController', async (): Promise<void> => {
...
+    it('should return an HTTP 200 and correct model', async () => {
+      const id = ctx.roles[0].id;
+
+      const res = await request(ctx.app)
+        .get(`/rbac/roles/${id}/all-users`)
```
JustSamuel, line 227:
```suggestion
        .get(`/rbac/roles/${id}/users`)
```

test/unit/controller/rbac-controller.ts
```diff
@@ -214,6 +219,35 @@ describe('RbacController', async (): Promise<void> => {
...
+        false,
+        true,
+      ).valid).to.be.true;
+    });
+
+    it('should return an HTTP 404 if role does not exist', async () => {
+      const id = ctx.roles.length + 1;
+
+      const res = await request(ctx.app)
+        .get(`/rbac/roles/${id}/all-users`)
```
JustSamuel, line 243:
```suggestion
        .get(`/rbac/roles/${id}/users`)
```

JustSamuel (approved):

PR #550, opened 2025-06-27
feat: check filter params for RBAC relation in get all transaction requests

JustSamuel:
# Description
Currently it is impossible to fetch the "createdBy" transactions of a user. This PR solves that issue by assigning the correct RBAC relation to a `GET /transactions` requests.

## Types of changes


src/controller/transaction-controller.ts
```diff
@@ -307,6 +307,43 @@ export default class TransactionController extends BaseController {
...
+        (fromId && await UserService.areInSameOrgan(userId, fromId)) ||
+          (toId && await UserService.areInSameOrgan(userId, toId)) ||
+          (createdById && await UserService.areInSameOrgan(userId, createdById))
+      ) {
+        return 'organ';
+      }
+
+      return 'all';
+    } catch (error) {
+      return 'all';
```
Yoronex, line 343: Maybe give the error to the logger as well, as we shouldn't end up here.

PR #569, opened 2025-07-07
fix: only delete users from gewisdbsync with 0 balance

JustSamuel:
# Description
Small fix, users should only be deleted if the balance is zero. Otherwise a write-off of something else is needed. 

## Types of changes


RubenLWF (approved):
LGTM, some checks are failing.

PR #640, opened 2025-11-17
feat: notification abstraction

SvenMokveldje:
Provide a general abstraction for notifications

# Description

## Types of changes

- Documentation improvement

---

## ✅ PR Checklist

### DIFF
--- a/src/controller/authentication-secure-controller.ts
+++ b/src/controller/authentication-secure-controller.ts
@@ -130,8 +130,12 @@ export default class AuthenticationSecureController extends BaseController {
     this.logger.trace('Refresh token for user', req.token.user.id);
 
     try {
-      const userOptions = UserService.getOptions({ id: req.token.user.id });
+      const userOptions = UserService.getOptions({ id: req.token.user.id, allowPos: true });
       const user = await User.findOne(userOptions);
+      if (!user) {
+        res.status(404).json('User not found.');
+        return;
+      }
       const token = await new AuthenticationService().getSaltedToken({
         user,
         context: {
@@ -220,8 +224,12 @@ export default class AuthenticationSecureController extends BaseController {
         return;
       }
 
-      const userOptions = UserService.getOptions({ id: req.token.user.id });
+      const userOptions = UserService.getOptions({ id: req.token.user.id, allowPos: true });
       const user = await User.findOne(userOptions);
+      if (!user) {
+        res.status(404).json('User not found.');
+        return;
+      }
       const token = await new AuthenticationService().getSaltedToken({
         user,
         context: {
--- a/src/service/user-service.ts
+++ b/src/service/user-service.ts
@@ -62,6 +62,7 @@ export interface UserFilterParameters {
   organId?: number,
   assignedRoleIds?: number | number[],
   pointOfSaleId?: number,
+  allowPos?: boolean,
 }
 
 export type FinancialMutationsFilterParams = TransactionFilterParameters & TransferFilterParameters;
@@ -210,8 +211,8 @@ export default class UserService {
     if (params.pointOfSaleId !== undefined) {
       where.pointOfSale = { id: params.pointOfSaleId };
       // When filtering by pointOfSaleId, we need to allow POINT_OF_SALE type users
-    } else if (!params.type) {
-      // Exclude POINT_OF_SALE type unless type filter is explicitly set
+    } else if (!params.type && !params.allowPos) {
+      // Exclude POINT_OF_SALE type unless type filter is explicitly set or allowPos is true
       where.type = Not(UserType.POINT_OF_SALE);
     }
 
--- a/test/unit/controller/authentication-secure-controller.ts
+++ b/test/unit/controller/authentication-secure-controller.ts
@@ -144,6 +144,31 @@ describe('AuthenticationSecureController', () => {
         true,
       ).valid).to.be.true;
     });
+
+    it('should handle refresh token request with POS token', async () => {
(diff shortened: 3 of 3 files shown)


src/mailer/messages/user-will-get-fined.ts
```diff
@@ -27,13 +27,19 @@
 import { Dinero } from 'dinero.js';
 import MailContentBuilder from './mail-content-builder';
 import MailMessage, { Language, MailLanguageMap } from '../mail-message';
+import { ParameterObject } from '../../notifications/notification-types';
 
-interface UserWillGetFinedOptions {
-  referenceDate: Date;
-  fine: Dinero;
-  balance: Dinero;
+export class UserWillGetFinedOptions extends ParameterObject {
```
JustSamuel: I don't think ParameterObject is a very good name hahah

src/notifications/channels/abstract-channel.ts
```diff
@@ -0,0 +1,48 @@
...
+  abstract send(user: User, content: TRendered): Promise<void>;
+
+  supports(type: string): boolean {
+    return type in this.templates;
+  }
+
+  getTemplate(type: string): TTemplate | undefined {
+    return this.templates[type];
+  }
+}
```
JustSamuel: I think it might be nice to ensure that the abstract class itself will always creat / save the NotificationLog entity?

that is you can have a structure in your abstract class as such
```ts
abstract class Abstract {
  abstract implementMe(): void

  common(): void {
    this.implementMe()
    // now do logging for example
    console.log('common finished')
  }
}

```

src/notifications/notifier.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  async notify<P extends TemplateOptions>(payload: NotificationPayload<P>): Promise<void> {
+    const notifyType = NotificationTypeRegistry.get<P>(payload.type);
+
+    if (!notifyType) {
+      this.logger.error(`Could not get notify type: ${payload.type}`);
+    }
+
+    const user = await User.findOne({ where: { id: payload.userId } });
```
JustSamuel: check if user exists

src/notifications/notifier.ts
```diff
@@ -0,0 +1,143 @@
...
+    if (channelsToUse.length === 0) {
+      await this.noChannelLog(user, payload.type);
+      return;
+    }
+
+    await Promise.allSettled(
+      channelsToUse.map(ch =>
+        this.sendViaChannel(ch as any, user, notifyType, payload.params),
+      ),
+    );
```
JustSamuel: Errors will be silent, maybe a .catch log?

JustSamuel (changes requested):
One big question remanining. What is the real use of "isMandetory"?

it seems unused. Does it note make more sense to do something like:

```ts
if (notifyType.isMandatory) {
  // Force send via default channel (EMAIL) regardless of preferences
  channelPrefs.push(NotificationChannels.EMAIL);
} else if (payload.overrideChannel) {
  channelPrefs.push(payload.overrideChannel);
} else {
  const userPrefs = await this.getPreferences(user, notifyType.type);
  channelPrefs.push(...userPrefs);
}
```

that way we can lessen the amount of overridechannel stuff?

src/notifications/notifier.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+    return this.instance;
+  }
+
+  async notify<P extends TemplateOptions>(payload: NotificationPayload<P>): Promise<void> {
+    const notifyType = NotificationTypeRegistry.get<P>(payload.type);
+
+    if (!notifyType) {
+      this.logger.error(`Could not get notify type: ${payload.type}`);
+    }
```
JustSamuel: Should we continue here? or error and return?

src/notifications/channels/mail-channel.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  readonly name = NotificationChannels.EMAIL;
+
+  async apply(template: EmailTemplate<any>, params: TemplateOptions): Promise<MailMessage<EmailTemplate<any>>> {
+    return template.build(params);
+  }
+
+  async send(user: User, email: MailMessage<EmailTemplate<any>>): Promise<void> {
+    await Mailer.getInstance().send(user, email);
+  }
```
JustSamuel: Regression: for some use cases we had that you could BCC the BACPM (FINANCIAL_RESPONSIBLE) so maybe a slight change to the email template or messages to include if the BACPM should be BCC? specifcially when closing accounts a BCC to the BACPM is nice.

src/notifications/notifier.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    const rendered = await channel.apply(template, params);
+    try {
+      await channel.log(user, notifyType.type);
+
+      await channel.send(user, rendered);
+    } catch (error) {
+      throw error;
+    }
```
JustSamuel: This will always log, also on failrue. Maybe you should only log _after_ send.

src/service/user-notification-preference-service.ts
```diff
@@ -0,0 +1,240 @@
...
+              userId: user.id,
+              type: combo.type,
+              channel: combo.channel,
+              enabled: false,
+            }),
+          ),
+        );
+      }
+    }
+  }
```
JustSamuel: Not sure how often we do this, but this seems _quite_ slow. Maybe a a single query or batch processing?

JustSamuel (approved):

PR #655, opened 2025-12-02
feat: SudoSOS Wrapped 2025

RubenLWF:
# Description

## Types of changes

---

## ✅ PR Checklist

### DIFF
--- a/src/maintenance.ts
+++ b/src/maintenance.ts
@@ -39,6 +39,7 @@ import { UserSyncService } from './service/sync/user/user-sync-service';
 import UserSyncManager from './service/sync/user/user-sync-manager';
 import GewisDBSyncService from './gewis/service/gewisdb-sync-service';
 import ServerSettingsStore from './server-settings/server-settings-store';
+import WrappedService from './service/wrapped-service';
 
 class MaintenanceApplication {
   logger: Logger;
@@ -95,6 +96,11 @@ async function performMaintenanceTasks(application: MaintenanceApplication): Pro
   application.logger.info('Updating balances...');
   await new BalanceService().updateBalances({});
   application.logger.info('Balances updated');
+
+  // Update wrapped
+  application.logger.info('Updating wrapped...');
+  await new WrappedService().updateWrapped({});
+  application.logger.info('Wrapped updated');
 
   // Setup user synchronization services based on environment variables
   const syncServices: UserSyncService[] = [];


src/service/wrapped-service.ts
```diff
@@ -0,0 +1,440 @@
...
+export default class WrappedService extends WithManager {
+  public async getWrappedForUser(userId: number): Promise<WrappedResponse | null> {
+    const entityManager = this.manager;
+    const query = `
+    SELECT w.* 
+    FROM wrapped w
+    INNER JOIN \`user\` u ON u.id = w.userId
+    WHERE w.userId = ?
+    LIMIT 1
+  `;
```
JustSamuel: why in the world would you need a query like this?

src/service/wrapped-service.ts
JustSamuel: I am quite sure there is 0 need for any rawsql. Maybe you need to a bit more creative, but take a look at the invoice service or others. 

Updating the raw JSON is maybe a bit more tricky, but that should also be fine.

src/entity/wrapped.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  @Column({
+    type: 'integer',
+  })
+  public transactionMaxAmount: number;
+
+  @Column({
+    type: 'text',
+  })
+  public transactionHeatmap: number[];
```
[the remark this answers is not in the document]
RubenLWF: Storing it in a json array appears to work just fine :)

src/controller/response/wrapped-response.ts
```diff
@@ -0,0 +1,50 @@
...
+export default interface WrappedResponse extends BaseResponse {
+  userId: number;
+  transactions: WrappedTransactions;
+}
+
+/**
+ * @typedef {object} WrappedTransactions
+ * @property {integer} count.required - The total number of wrapped transactions
+ * @property {integer} count2024.required - The number of wrapped transactions in 2024
+ * @property {number} transactionPercentile.required - The percentile of the user based on wrapped transactions
```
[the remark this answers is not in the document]
RubenLWF: While I do think that these statistics could be interesting to implement, I would rather not implement it and make sure that it is finished before the deadline (Since this will also bring more frontend work with it)

src/service/wrapped-service.ts
```diff
@@ -0,0 +1,333 @@
...
+import SubTransactionRow from '../entity/transactions/sub-transaction-row';
+import ProductRevision from '../entity/product/product-revision';
+import { Between, In } from 'typeorm';
+
+export interface UpdateWrappedParameters {
+  ids?: number[],
+}
+
+export default class WrappedService extends WithManager {
+  public async getWrappedForUser(userId: number): Promise<WrappedResponse | null> {
```
JustSamuel: Also, this is not really your fault, but I try to work more towards this goal:
https://github.com/GEWIS/sudosos-backend/issues/153

and that returning a `Wrapped` and having the controller turn the `Wrapped` into a `WrappedResponse` is prefered since this makes the WrappedService more flexible.

src/service/wrapped-service.ts
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    const res = new Map<number, number>();
+    for (const entry of values) {
+      const less = lessCountMap.get(entry.value) ?? 0;
+      const pct = 100 * (1 - (less / total));
+      res.set(entry.id, Number(pct.toFixed(2)));
+    }
+    return res;
+  }
```
JustSamuel: Some comments could help, specifically some JSDOC (here and at multiple places) since it can be hard to follow what the intention of a function is.

src/service/wrapped-service.ts
```diff
@@ -0,0 +1,333 @@
...
+
+  public async updateWrapped(params: UpdateWrappedParameters = {}) {
+    let users : User[];
+    let rows : Wrapped[] = [];
+
+    const syncedFrom = new Date(2025, 0, 1, 0, 0, 0);
```
JustSamuel: I dont like hardcoding this date. How would this work next year?

src/service/wrapped-service.ts
```diff
@@ -0,0 +1,450 @@
...
+  }
+
+  /**
+   * Compute percentile rank for each entry in the input list.
+   *
+   * This function calculates, for each entry, the percentile that its `value`
+   * falls into among all provided values.
+   *
+   * Example: if 2 out of 10 values are strictly less than a value, its percentile
+   * becomes 100 * (1 - 2/10) = 80.00.
```
JustSamuel: A bit heavy on the vibecoded comments, but its fine.

PR #692, opened 2026-01-12
ci: bump node version in Dockerfile

JustSamuel:
# Description
Bump from 18 to 22

## Types of changes


JustSamuel:
Fixed @RubenLWF, it now uses alpine again.

PR #729, opened 2026-02-02
feat: implement bullMQ

SvenMokveldje:
This adds bullMQ to the SudoSOS repository and immediatly implements it for the mail services.

# Description

Add 2 new environment variables:
**REDIS_HOST**
**REDIS_PORT**

## Types of changes

---

## ✅ PR Checklist

### DIFF
--- a/src/mailer/mailer.ts
+++ b/src/mailer/mailer.ts
@@ -60,6 +60,7 @@ export default class Mailer {
     this.mailQueue = new Queue('mail-queue', {
       connection: redisConnection as unknown as ConnectionOptions,
     });
+    Mailer.instance = this;
   }
 
   static getInstance(): Mailer {
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,6 +69,7 @@
         "@types/swagger-ui-express": "^4.1.8",
         "@types/uuid": "^10.0.0",
         "@types/validator": "^13.15.10",
+        "@types/ws": "^8.18.1",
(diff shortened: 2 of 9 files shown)


src/mailer/mailer.ts
RubenLWF: Suggestion:
Would it not make sense to, rather than defining this in the mailer, define an application-wide connection to Redis in, for example, `index.ts`?

We could still define the mailqueue here, but this would allow us to more easily use Redis in multiple places in the future.

src/workers/mail-worker.ts
```diff
@@ -0,0 +1,67 @@
...
+        );
+
+        throw error;
+      }
+    },
+    {
+      connection: {
+        host: process.env.REDIS_HOST || '127.0.0.1',
+        port: Number(process.env.REDIS_PORT) || 6379,
+      },
```
RubenLWF: Note:
Having the application wide connection to Redis, would make it so that you could use the already existing connection here, rather than having to define a new one.

src/index.ts
```diff
@@ -301,6 +306,9 @@ export default async function createApp(): Promise<Application> {
 
   webSocketService.initiateWebSocket();
 
+  application.workers = [];
+  application.workers.push(startMailWorker());
```
RubenLWF: Suggestion:
```suggestion
  application.workers = [startMailWorker()];
```

RubenLWF (changes requested):
Nice work!

I have left some comments/questions/things to consider.

PR #782, opened 2026-03-06
Migrate test suite to use production roles instead.

### LINKED ISSUE #166: Production RBAC roles and permissions are not tested

JustSamuel:
# Description
Current setup used ad-hoc roles, this migrates everything to use production roles.

Closes #166 

## Types of changes

### DIFF
--- a/test/helpers/user-factory.ts
+++ b/test/helpers/user-factory.ts
@@ -43,9 +43,20 @@ async function getProductionRoleNames(userType: UserType): Promise<string[]> {
     where: { systemDefault: true },
     relations: { roleUserTypes: true },
   });
-  return roles
+  const roleNames = roles
     .filter((r) => r.roleUserTypes.some((rut) => rut.userType === userType))
     .map((r) => r.name);
+
+  // Guard against unsynchronized default roles: an empty result likely means
+  // ensureProductionRoles()/DefaultRoles.synchronize() has not been called.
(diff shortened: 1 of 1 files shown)


src/rbac/default-roles.ts
```diff
@@ -204,6 +204,11 @@ export default class DefaultRoles {
         Balance: admin,
         Banner: admin,
         Container: admin,
+        Event: admin,
+        EventAnswer: {
+          ...admin,
+          assign: { all: star },
+        },
```
RubenLWF: Question:

Can we already delete the events code? Or do we still need to support this for some time?

test/helpers/user-factory.ts
```diff
@@ -21,6 +21,67 @@
...
+    );
+  }
+
+  return roleNames;
+}
+
+/**
+ * Build a JsonWebToken for a user using their production roles.
+ */
+export async function tokenFor(
```
RubenLWF: suggestion:

Since the `signTokenFor` function already exists, and we never need an unsigned token, make this function private.

PR #821, opened 2026-03-25
Add summary endpoint and CREDIT_INVOICE type

JustSamuel:
# Description
Adds a `summary` endpoint to transfers, which combines all the aggregations. Also adds `CREDIT_INVOICE` as an transfer type.

## Types of changes

=

### DIFF
--- a/src/controller/transfer-controller.ts
+++ b/src/controller/transfer-controller.ts
@@ -29,7 +29,7 @@ import log4js, { Logger } from 'log4js';
 import BaseController, { BaseControllerOptions } from './base-controller';
 import Policy from './policy';
 import { RequestWithToken } from '../middleware/token-middleware';
-import TransferService, { parseGetTransferAggregateFilters, parseGetTransferFilters } from '../service/transfer-service';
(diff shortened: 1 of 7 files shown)


src/controller/transfer-controller.ts
```diff
@@ -143,6 +149,53 @@ export default class TransferController extends BaseController {
...
+   * @summary Returns an aggregate breakdown of transfers for every category plus an overall total
+   * @operationId getTransferSummary
+   * @tags transfers - Operations of transfer controller
+   * @security JWT
+   * @param {string} fromDate.query - Start date for selected transfers (inclusive)
+   * @param {string} tillDate.query - End date for selected transfers (exclusive)
+   * @param {integer} fromId.query - Filter transfers from this user ID
+   * @param {integer} toId.query - Filter transfers to this user ID
+   * @return {TransferSummaryResponse} 200 - Per-category aggregate sums and counts
+   * @return {string} 400 - Validation error
```
CodeNamedRobin: Can this also not return 404 if the user is not found? Seems weird to turn that into a bad request.

Issue #780, opened 2026-03-05
[Refactor]: Centralize environment variables into a typed Config singleton


JustSamuel:
### What would you like?

A typed singleton `Config` class that centralizes all ~60 environment variables. Instead of scattered `process.env.X` calls with inconsistent parsing and fallback logic, all env access goes through one object loaded once at startup.

### Why is this needed?

- **No startup validation** — missing required vars (e.g. `STRIPE_PRIVATE_KEY`) only crash when that code path is first hit at runtime, not at boot.
- **Inconsistent parsing** — some vars use `parseInt()` with NaN fallbacks, others use `Number()`, others use `||` vs `??`. Boolean checks mix `=== 'true'` patterns.
- **Duplicated defaults** — fallback values like `BCRYPT_ROUNDS = 12` or `REDIS_PORT = 6379` are hardcoded inline wherever they're used.
- **Hard to discover** — no single place to see what env vars the app needs. You have to grep across 40+ source files.

### How could it be implemented?

A `src/config.ts` singleton with typed properties, loaded once via `dotenv` + parsing:

```typescript
class Config {
  private static instance: Config;

  // Groups
  readonly database: { connection: string; host: string; port: number; ... };
  readonly redis: { host: string; port: number; connectTimeoutMs: number };
  readonly stripe: { privateKey?: string; publicKey?: string; ... };
  readonly smtp: { host?: string; port?: number; ... };
  // ...

  private constructor() { /* parse + validate all process.env once */ }

  static get(): Config {
    if (!this.instance) this.instance = new Config();
    return this.instance;
  }

  // For test isolation
  static reset(): void { this.instance = undefined; }
}
```

**Refactor scope:** Medium — mechanically replace `process.env.X` across ~40-50 files. No database or API changes. Low risk since it's the same values, just centralized.

**Test impact:** `test/setup.ts` already sets env vars before anything runs, so the singleton would pick them up on first access. A `Config.reset()` method handles test isolation.

### Other information

Current inventory: ~60 unique env vars, ~24 with hardcoded defaults, 2 explicitly required (`NAME`, `JWT_KEY_PATH`), ~15 conditionally required (LDAP, Stripe, SMTP, GEWIS groups).

PR #833, opened 2026-03-31
Add k8s kustomize setup

JustSamuel:
mooi


JustSamuel:
Might also be nice to combine this with #780

Issue #116, opened 2023-12-26
Migrate all request validators to custom `swaggar-model-async-validator`


JustSamuel:
I created a fork of the `swagger-model-validator` package that we use. This fork (bumps all the dependencies) and allows for custom validator functions to be async. This means we can replace my over complicated custom validation logic with something that is way clearer.

Migration should not be to hard. [Here](https://github.com/GEWIS/swagger-model-async-validator) is how to add custom validators. So we just have to rewrite, and mostly remove code. 

### Progress
- [ ] container-request-spec.ts
- [ ] duration-spec.ts
- [ ] general-validators.ts
- [ ] invoice-request-spec.ts
- [ ] point-of-sale-request-spec.ts
- [ ] product-request-spec.ts
- [ ] string-spec.ts
- [ ] update-local-request-spec.ts
- [ ] update-nfc-request-spec.ts
- [ ] update-pin-request-spec.ts
- [ ] user-request-spec.ts
- [ ] validation-errors.ts

PR #834, opened 2026-04-01
Lift inactive admin cost validators into AsyncValidatorMiddleware

JustSamuel:
Partial #116 (InactiveAdministrativeCostController items).

## Summary
- Register `CreateInactiveAdministrativeCostRequest` and `HandoutInactiveAdministrativeCostsRequest` specs with `globalAsyncValidatorRegistry` in the `InactiveAdministrativeCostController` constructor
- Export `createInactiveAdministrativeCostRequestSpec` and `handoutInactiveAdministrativeCostsRequestSpec` as factory functions so the registry produces a fresh spec per request
- Remove all inline `isFail`/`verifyValidUserId` checks from `createInactiveAdministrativeCost` handler
- Remove inline userIds array validation from `notifyInactiveUsers` and `handoutInactiveAdministrativeCost` handlers
- Update controller tests to assert the `{ valid: false, errors[] }` response shape instead of bare strings

## Test plan
- [x] `tsc --noEmit` passes
- [x] `npm run lint` passes
- [x] All 44 controller tests pass
- [x] Validation tests updated to assert `{ valid: false, errors[] }` shape


RubenLWF (approved):
Change looks good.

Please have a look at the Copilot comment and fix the merge conflict.

PR #835, opened 2026-04-01
Centralize env config setup

### LINKED ISSUE #780: [Refactor]: Centralize environment variables into a typed Config singleton

JustSamuel:
# Description
Move everything env related to a single source of truth with sensible defaults.

Closes #780

## Types of changes
- Refactor

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,0 +1,366 @@
+/**
+ *  SudoSOS back-end API service.
+ *  Copyright (C) 2026 Study association GEWIS
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as published
+ *  by the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ *  @license
+ */
+
+import dotenv from 'dotenv';
+
+dotenv.config();
+
+export type DatabaseConnection = 'sqlite' | 'postgres' | 'mariadb' | 'mysql';
+const VALID_DATABASE_CONNECTIONS: readonly string[] = ['sqlite', 'postgres', 'mariadb', 'mysql'];
+
+export type StorageMethod = 'disk';
+const VALID_STORAGE_METHODS: readonly string[] = ['disk'];
+
+function getOptionalString(name: string): string | undefined {
+  const value = process.env[name];
+  if (value == null) return undefined;
+
+  const trimmed = value.trim();
+  // Node.js stringifies `process.env.X = undefined` as "undefined"
+  return (trimmed === '' || trimmed === 'undefined') ? undefined : trimmed;
+}
+
+function getInteger(
+  name: string,
+  fallback: number,
+): number {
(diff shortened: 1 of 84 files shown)


src/helpers/pagination.ts
```diff
@@ -25,14 +25,15 @@
  */
 
 import { Request } from 'express';
+import Config from '../config';
 
-export const PAGINATION_DEFAULT = 25;
+export const PAGINATION_DEFAULT = 20;
```
RubenLWF: question:

why this change?

src/config.ts
```diff
@@ -0,0 +1,368 @@
...
+  return Number.isNaN(parsed) ? fallback : parsed;
+}
+
+function getBoolean(name: string, fallback: boolean = false): boolean {
+  const rawValue = getOptionalString(name);
+  if (rawValue == null) return fallback;
+
+  return rawValue === 'true';
+}
+
```
RubenLWF, line 57: suggestion:

remove this double empty line since we do not do this anywhere else

src/config.ts
```diff
@@ -0,0 +1,368 @@
...
+function getOptionalString(name: string): string | undefined {
+  const value = process.env[name];
+  if (value == null) return undefined;
+
+  const trimmed = value.trim();
+  // Node.js stringifies `process.env.X = undefined` as "undefined"
+  return (trimmed === '' || trimmed === 'undefined') ? undefined : trimmed;
+}
+
+
```
RubenLWF: suggestion:

remove this double empty line since we do not do this anywhere else

RubenLWF (changes requested):

src/helpers/pagination.ts
JustSamuel: The constant was accidentally changed to 20 to match the hardcoded fallback in `Config` — but that was the wrong direction. Reverted `PAGINATION_DEFAULT` back to 25 and updated the fallback in `config.ts` to 25 as well so they're consistent.

PR #836, opened 2026-04-01
fix: incorrect enum definition

CodeNamedRobin:
Changes input param for the get all balances function from Array<string|number> to Array<string>

# Description
Makes sure enum is correctly defined in the balance controller JSDoc.

## Types of changes

## ✅ PR Checklist

- [X] **Test Coverage**: New functionality has appropriate test coverage and all tests pass (`npm run test`)
- [X] **Documentation**: New functionality is documented with TypeDoc comments and API documentation is updated


RubenLWF (approved):

JustSamuel (approved):
I think numbers are still supported?

CodeNamedRobin:
> I think numbers are still supported?

From what I can see in the backend, numbers are not supported. This is also reflected in when I put in a "1" in the first box, it just converts it to a string and says "type not valid"

PR #843, opened 2026-04-03
fix: transfers double counting in aggregate

CodeNamedRobin:
Checks if all categories are NULL when querying manual transfers

# Description
Fixes a bug where all transfers would be counted as manually created or deleted in the transfer aggregate.
## Types of changes
---

## ✅ PR Checklist

- [X] **Test Coverage**: New functionality has appropriate test coverage and all tests pass (`npm run test`)
- [X] **Documentation**: New functionality is documented with TypeDoc comments and API documentation is updated

### DIFF
--- a/src/service/transfer-service.ts
+++ b/src/service/transfer-service.ts
@@ -295,15 +295,25 @@ export default class TransferService extends WithManager {
       [TransferCategory.INACTIVE_ADMINISTRATIVE_COST]: 'inactiveAdministrativeCost',
     };
 
+    const excludeAllCategories = (q: typeof query) => {
+      for (const category of Object.values(categoryRelationMap)) {
+        q = q.leftJoin(`transfer.${category}`, category)
+          .andWhere(`${category}.id IS NULL`);
+      }
+      return q;
+    };
+
     let query = this.manager.createQueryBuilder(Transfer, 'transfer');
 
     if (filters.category !== undefined) {
       switch (filters.category) {
         case TransferCategory.MANUAL_CREATION:
-          query = query.andWhere('transfer.fromId IS NULL');
+          query = excludeAllCategories(query)
+            .andWhere('transfer.fromId IS NULL');
           break;
         case TransferCategory.MANUAL_DELETION:
-          query = query.andWhere('transfer.toId IS NULL');
+          query = excludeAllCategories(query)
+            .andWhere('transfer.toId IS NULL');
           break;
         default: {
           const rel = categoryRelationMap[filters.category];
--- a/test/unit/service/transfer-service.ts
+++ b/test/unit/service/transfer-service.ts
@@ -524,7 +524,25 @@ describe('TransferService', async (): Promise<void> => {
 
     it('should return all transfers with null fromId for the MANUAL_CREATION category', async () => {
       const manualCreations = await Transfer.createQueryBuilder('transfer')
-        .where('transfer.fromId IS NULL')
+        .leftJoin('transfer.deposit', 'deposit')
+        .leftJoin('transfer.payoutRequest', 'payoutRequest')
+        .leftJoin('transfer.sellerPayout', 'sellerPayout')
+        .leftJoin('transfer.invoice', 'invoice')
+        .leftJoin('transfer.creditInvoice', 'creditInvoice')
+        .leftJoin('transfer.fine', 'fine')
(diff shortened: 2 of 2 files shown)


src/service/transfer-service.ts
```diff
@@ -299,11 +299,18 @@ export default class TransferService extends WithManager {
...
+      // Transfer is only manually created if no other category can be assigned to it
         case TransferCategory.MANUAL_CREATION:
-          query = query.andWhere('transfer.fromId IS NULL');
-          break;
         case TransferCategory.MANUAL_DELETION:
-          query = query.andWhere('transfer.toId IS NULL');
+          for (const category of Object.values(categoryRelationMap)) {
+            query = query.andWhere(`transfer.${category} IS NULL`);
+          }
+          query = query.andWhere('invoice.creditTransferId IS NULL');
```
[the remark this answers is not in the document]
CodeNamedRobin: Applied suggestion in latest commit

src/service/transfer-service.ts
```diff
@@ -299,11 +299,18 @@ export default class TransferService extends WithManager {
 
     if (filters.category !== undefined) {
       switch (filters.category) {
+      // Transfer is only manually created if no other category can be assigned to it
         case TransferCategory.MANUAL_CREATION:
```
RubenLWF, line 310: question:

If we do nothing in case of `MANUAL_CREATION` this case can be removed altogether right?

src/service/transfer-service.ts
```diff
@@ -299,11 +299,18 @@ export default class TransferService extends WithManager {
...
-          query = query.andWhere('transfer.fromId IS NULL');
-          break;
         case TransferCategory.MANUAL_DELETION:
-          query = query.andWhere('transfer.toId IS NULL');
+          for (const category of Object.values(categoryRelationMap)) {
+            query = query.leftJoin(`transfer.${category}`, category)
+              .andWhere(`${category}.id IS NULL`);
+          }
+          if (filters.category === TransferCategory.MANUAL_CREATION) {
```
RubenLWF: question:

This can never be the case right? Since we are inside the `MANUAL_DELETION` case.

RubenLWF (changes requested):

src/service/transfer-service.ts
CodeNamedRobin: We don't do nothing in the `MANUAL_CREATION` case, we fallthrough to the other case. This syntax means boils down to 'do the same thing for both manual creation and deletion'. This also means that we can still distinguish between manually created and deleted in this case. I found this reddit thread to be helpful to understand the fallthrough logic.
https://www.reddit.com/r/javascript/comments/7pbmqh/what_is_fallthrough_in_javasript/

src/service/transfer-service.ts
CodeNamedRobin: See previous comment

RubenLWF: In that case, since there are still different actions happening for the 2 cases, I would prefer to keep it in 2 separate cases, rather than 1 case with an if statement inside it

src/service/transfer-service.ts
```diff
@@ -299,11 +299,18 @@ export default class TransferService extends WithManager {
...
+            query = query.andWhere('transfer.fromId IS NULL');
+          } else {
+            query = query.andWhere('transfer.toId IS NULL');
+          }
```
RubenLWF: please ignore the indentation

``` suggestion
        case TransferCategory.MANUAL_CREATION:
          for (const category of Object.values(categoryRelationMap)) {
            query = query.leftJoin(`transfer.${category}`, category)
              .andWhere(`${category}.id IS NULL`);
          }
          query = query.andWhere('transfer.fromId IS NULL');
          break;
        case TransferCategory.MANUAL_DELETION:
          for (const category of Object.values(categoryRelationMap)) {
            query = query.leftJoin(`transfer.${category}`, category)
              .andWhere(`${category}.id IS NULL`);
          }
            query = query.andWhere('transfer.toId IS NULL');
          break;
```

RubenLWF (changes requested):

src/service/transfer-service.ts
CodeNamedRobin: Extracted the common function into `excludeAllCategories` and created two separate cases for `MANUAL_CREATION` and `MANUAL_DELETION`

RubenLWF (approved):

PR #853, opened 2026-04-04
ci: remove namespace definitions from k8s overlays

JustSamuel:
## Summary
- CBC controls namespace creation — SudoSOS should not define \`Namespace\` objects or stamp a \`namespace:\` field in overlays
- Removed \`namespace.yaml\` from both overlays
- Removed \`namespace:\` field from both overlay \`kustomization.yaml\` files
- Namespaces will be injected via \`spec.targetNamespace\` in the Flux Kustomization resources in k8s-infra

## Test plan
- [x] No new tests needed (infra config only)
- [x] No DB migration


RubenLWF (approved):
hype

PR #856, opened 2026-04-04
ci: replace nginx-ingress annotation with Traefik equivalent

JustSamuel:
## Summary
- Removes `nginx.ingress.kubernetes.io/proxy-body-size: "50m"` which is silently ignored by Traefik
- Adds `traefik.ingress.kubernetes.io/router.entrypoints: websecure` so Traefik routes via HTTPS
- File upload size limits are enforced by the internal nginx pod, not the ingress controller

## Test plan
- [x] No new tests needed (k8s config only)
- [x] No DB migration

### DIFF
--- a/k8s/base/nginx/ingress.yaml
+++ b/k8s/base/nginx/ingress.yaml
@@ -2,8 +2,6 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: sudosos
-  annotations:
-    traefik.ingress.kubernetes.io/router.entrypoints: web
 spec:
   rules:
     - host: sudosos.gewis.nl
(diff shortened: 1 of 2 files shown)


k8s/base/nginx/ingress.yaml
```diff
@@ -3,7 +3,7 @@ kind: Ingress
 metadata:
   name: sudosos
   annotations:
-    nginx.ingress.kubernetes.io/proxy-body-size: "50m"
```
LuukBlankenstijn, line 6: This is the opposite of what the description says, and also redundant. All routing in the cluster is websecure by default

LuukBlankenstijn (changes requested):

LuukBlankenstijn (approved):

PR #875, opened 2026-04-10
Switch backend package manager from npm to pnpm

JustSamuel:
## Summary

- Replace `package-lock.json` with `pnpm-lock.yaml` (generated with pnpm 10.33.0)
- Add `pnpm.overrides` for ioredis (replaces npm `overrides`; keeps the `$ioredis` self-reference syntax which pnpm also supports)
- Add `pnpm.onlyBuiltDependencies` to explicitly allow native modules to run their build scripts: `bcrypt`, `sqlite3`, `esbuild`, `msgpackr-extract`, `vue-demi` (required by pnpm 10's security model)
- Update all `npm run X` references in `package.json` scripts to `pnpm run X`
- Update `Dockerfile`: installs pnpm via `npm install -g pnpm`, uses `pnpm install --frozen-lockfile` / `pnpm --prod`
- Update `Dockerfile-docs`: same pnpm setup in both build stages
- Update CI workflows — container jobs (`build.yml`, `coverage-baseline.yml`): `npm install -g pnpm` then use pnpm; non-container jobs (`publish-client.yml`, `release.yml`): use `pnpm/action-setup@v4`
- All workflows hash `pnpm-lock.yaml` for node_modules caching
- The `client/` subdirectory retains npm (generated artefact with its own lock file)
- `node_modules` is preserved

## Test plan

- [x] `pnpm install` generates a valid `pnpm-lock.yaml` and populates `node_modules`
- [x] `npm run lint` passes with no errors
- [x] CI workflows updated to install pnpm and call `pnpm install --frozen-lockfile`
- [x] Dockerfiles install pnpm before running build commands
- [x] `pnpm.onlyBuiltDependencies` ensures native modules compile correctly
- [x] No DB migration needed


.github/workflows/build.yml
```diff
@@ -21,32 +21,34 @@ jobs:
       image: node:24
     steps:
       - uses: actions/checkout@v4
+      - run: npm install -g pnpm@10.33.0
```
RubenLWF, line 24: question:

Should we "hardcode" this pnpm version in all pipelines?

README.md
RubenLWF, line 1: We should add pnpm to the list of prerequisites as well.

RubenLWF (changes requested):
Looking good! 😎

.github/workflows/build.yml
JustSamuel: Yes, not pining dependencies is the new smoking.

RubenLWF (approved):
🦫

PR #884, opened 2026-04-17
Migrate test runner from Mocha to Vitest

### LINKED ISSUE #353: [Chore] Update `chai`
### What would you like?

As discussed in https://github.com/GEWIS/sudosos-backend/pull/305, there is a problem with updating `chai`, the package we use for our test suite. At some point, we will have to update it or replace it with some other alternative.

### Why is this needed?

- The newest version of chai only works with ESM code. https://github.com/chaijs/chai/issues/1568
- TypeORM / es-build only compile to `commonjs`.
This is a pretty annoying dependency issue that I was not able to fix at this point in time. We are still on an old version and it works for now, but who knows for how long?

### How could it be implemented?

_No response_

### Other information

_No response_

JustSamuel:
# Description

Migrates the backend test runner from **Mocha + nyc** to **Vitest + @vitest/coverage-v8** (with SWC transform), unblocking the chai v5 upgrade (#353). Vitest supports ESM natively and sidesteps the CJS/ESM interop that blocked #305.

The migration is mechanical in most files — Mocha root-hooks → top-level `beforeAll`/`beforeEach`/`afterEach`/`afterAll`, `this.skip()` → `ctx.skip()`, `require()` → ESM imports — but several pockets needed rethinking:

## What needed real work

- **ESM-unstubbable bindings**: `sinon.stub(namespace, 'exportName')` fails under ESM because named exports are read-only live bindings. The `transaction.html` tests were rewritten to assert on HTML output directly instead of stubbing `createBasePdf`, and `invoice-controller` now stubs `Invoice.prototype.validatePdfHash` instead of `hash.hashJSON`.
- **SWC enum metadata**: SWC emits `Object` for TS enum columns in decorator metadata, not `String`/`Number` like tsc. Entity columns backed by enums now declare an explicit `type: 'varchar'` / `type: 'int'` so TypeORM picks the right column type.
- **Nested `it()` pattern**: `testValidationOnRoute` in `product-controller` was called inside outer `it()` blocks. Mocha silently dropped the inner registrations (so the tests never actually ran); Vitest rejects the pattern. Moving to `describe()` wrappers exposed pre-existing assertion mismatches, which are also fixed here.
- **Config vs process.env**: the `ldap-sync-service` test expected `process.env.LDAP_SERVICE_ACCOUNT_FILTER` but the source reads `Config.get().ldap.serviceAccountFilter`, which normalizes empty strings to `undefined`. Test now compares against `Config.get()`.
- **Pre-existing bugs Mocha masked**: fixed 7 unhandled rejections Vitest surfaces — missing `return` after 400 responses in `vat-group-controller`, a triple-no-op async `.every` assertion in `container-controller`, a dangling rejected promise in `token-handler`, and a `beforeAll(() => async () => {...})` typo in `vat-group-service` where the outer arrow returned the async fn instead of running it.

## CI

Matches the existing two-matrix strategy:

- **mariadb**: runs `pnpm coverage-ci-migrate` — Vitest against TS sources via SWC.
- **better-sqlite3**: runs `pnpm test-js-ci` — Vitest against the compiled `out/test/**/*.js` (validates the build artifact actually runs, same philosophy as the old `test-js` script). Uses a separate `vitest.config.js.mts` with `fileParallelism: false` to avoid CJS `Queue.prototype` state-leaking across test files.

Config files use `.mts` extension to silence Vite's CJS-Node-API deprecation warning.

- Fixes #353
- Relates to #305

## Types of changes

- Refactoring _(test infrastructure)_

---

## ✅ PR Checklist

- [x] **Test Coverage**: 2422 tests pass / 27 skipped under Vitest (same skip conditions as before — LDAP tests without external services, Stripe tests without test keys). Both TS and compiled-JS paths green with 0 unhandled rejections.
- [x] **Documentation**: No API changes — test infrastructure only.
- [x] **Database Changes**: No schema changes (entity column-type annotations are metadata hints; actual DB column types are unchanged).

## 🔗 Additional Notes

Verified locally:
- `pnpm test-ci` → 2422 passed / 27 skipped / 103 test files, exit 0
- `pnpm test-js-ci` → 2422 passed / 27 skipped / 103 test files, exit 0
- `pnpm lint` → clean
- `pnpm build` → clean


src/entity/stripe/stripe-payment-intent-status.ts
```diff
@@ -43,6 +43,6 @@ export default class StripePaymentIntentStatus extends BaseEntity {
   @ManyToOne(() => StripePaymentIntent, (intent) => intent.paymentIntentStatuses, { nullable: false })
   public stripePaymentIntent: StripePaymentIntent;
 
-  @Column()
+  @Column({ type: 'int' })
```
RubenLWF, line 46: question:

Is there a reason we are not consistent in whether an enum is an `int` or a `varchar`?

test/unit/html/transaction.html.ts
```diff
@@ -169,7 +152,7 @@ describe('transaction.html', () => {
       const html = createTransactionPdf(options);
 
       expect(html).to.be.a('string');
-      expect(createBasePdfStub).to.have.been.calledOnce;
+      expect(html).to.include('999');
```
RubenLWF: question:

Where did this check come from?

test/unit/service/websocket-service.ts
```diff
@@ -28,6 +28,13 @@ import TokenHandler from '../../../src/authentication/token-handler';
 import RoleManager from '../../../src/rbac/role-manager';
 import User from '../../../src/entity/user/user';
 
+const asPromise = (
+  body: (done: (err?: any) => void) => void,
+): (() => Promise<void>) => () => new Promise<void>((resolve, reject) => {
+  const done = (err?: any): void => (err ? reject(err) : resolve());
+  body(done);
+});
+
```
RubenLWF, line 40: question:

why do we need this?

test/unit/database.ts
```diff
@@ -53,6 +52,7 @@ describe('Database', async (): Promise<void> => {
         'bigint': ['bigint'],
         'float': ['real', 'float'],
         'integer': ['integer', 'int'],
+        'int': ['integer', 'int'],
```
RubenLWF, line 55: question:

do we use thes interchangibly?

if so, should we?

test/setup.ts
```diff
@@ -100,6 +100,7 @@ if (!process.env.TYPEORM_CONNECTION || (process.env.TYPEORM_CONNECTION === 'bett
   process.env.TYPEORM_SYNCHRONIZE = 'true';
 }
 
+
```
RubenLWF: nitpick:

this can be removed

RubenLWF (changes requested):
🦫 

You are removing quite some timeouts from several tests, do we not need these anymore?

src/entity/stripe/stripe-payment-intent-status.ts
JustSamuel: The column type mirrors the enum's underlying value type:

- Numeric-valued enums → `int` (here: `StripePaymentIntentState = 1|2|3|4`, also `InvoiceState = 1|2|3|4`)
- String-valued enums → `varchar` (e.g. `PayoutRequestState`, `Availability`, `UserType`, `NotificationChannels`)

So it is consistent — by enum value type. The explicit annotations are only needed because SWC emits `Object` for TS enums in decorator metadata (unlike `tsc` which emits `String`/`Number`), so TypeORM can no longer infer it automatically.

test/unit/html/transaction.html.ts
JustSamuel: Fair question — it was a leftover arbitrary smoke check from when I replaced the stub-based assertions. Changed it to `expect(html).to.include('0,00')`, which actually verifies the empty-cart case (no items → subtotal/total render as zero).

test/unit/service/websocket-service.ts
JustSamuel: It bridges socket.io's callback-style flow into Promises. The old tests used Mocha's `done` callback (e.g. `clientSocket.on('event', () => done())`), which Vitest doesn't support — hooks/tests must return a Promise. Rather than rewrite all 19 tests into async/await + manual Promise construction, the helper wraps the existing callback pattern.

Added a comment in the file making this explicit.

test/unit/database.ts
JustSamuel: Yes — they're aliases. TypeORM accepts both `type: 'int'` and `type: 'integer'` and treats them identically, and our entities happen to use both (`balance.ts` declares `'integer'`; the enums I annotated in this PR use `'int'`). The two map keys are there so the test's lookup works regardless of which alias an entity uses — both point at the same set of acceptable DB column types (`['integer', 'int']`).

Agreed we should pick one going forward, but normalizing the entity side felt out of scope for the test-runner migration. Happy to open a follow-up.

test/setup.ts
JustSamuel: Addressed — removed the extra blank line.

JustSamuel:
> You are removing quite some timeouts from several tests, do we not need these anymore?

They're now covered by a global default. `vitest.config.mts` / `vitest.config.js.mts` set:

```ts
testTimeout: 50000,
hookTimeout: 50000,
```

Every `this.timeout(…)` I removed was either `50000` (same as the new global — so fully redundant) or the one-off `this.timeout(5000)` in `point-of-sale-service.ts`, which now runs with the 50s global — a relaxation, not a regression. No test loses timeout coverage.

Using a single global rather than per-hook settings also matches Vitest's idiomatic way of configuring this — Mocha's `this.timeout` on the hook context isn't available the same way (`this` there would be the Vitest test context instead).

RubenLWF (approved):

Issue #139, opened 2024-01-20
Introduce possibility for final 'invoice'


rinkp:
Currently, there is no way to settle a positive or negative account balance after the user has been disabled. A final ~invoice~ payment overview shall be an option for these cases. 
It shall not be an invoice, because we already computed VAT during the transaction in SudoSOS since these users are not invoice accounts.

JustSamuel:
This depends on #639 (PaymentLink entity) being implemented first — the payment link is the mechanism that would actually deliver the settlement to the user's account. Suggested order: #639 → #139 / #197.

PR #890, opened 2026-04-24
Add PaymentRequest primitive for shareable top-up links

### LINKED ISSUE #639: [Feature]: PaymentRequest entity
### What would you like?

An entity and flow allowing for unauthenticated, predetermined top-ups, delivered to the user as a shareable "payment link". You receive a link and you pay the fixed amount that was decided on creation.

### Why is this needed?

This is the first step towards implementing [Automagic Invoices (#197)](https://github.com/GEWIS/sudosos-backend/issues/197) and can also be used in combination with the [final "invoice" (#139)](https://github.com/GEWIS/sudosos-backend/issues/139). Ship this as a standalone primitive first, then wire those use cases on top.

## Overview

Enable users to top-up via shareable links without requiring authentication. The entity in the domain is a `PaymentRequest` (what it *is*); the "payment link" is how it is delivered to the user. This issue covers the primitive only — invoice coupling and settlement-for-disabled-users are tracked in their own issues.

## Current State

- **Top-up flow**: `POST /stripe/deposit` requires JWT authentication, creates `StripePaymentIntent` + `StripeDeposit`, webhook creates `Transfer` on success.

## Scope of this issue

**In scope (v1):**
- `PaymentRequest` entity + service + controller.
- Public endpoints to view and pay a request.
- Authenticated endpoints to create, list, and cancel.
- Integration with the existing Stripe flow (`StripePaymentIntent` → `StripeDeposit` → `Transfer`).

**Out of scope (tracked elsewhere):**
- Automatic creation of `PaymentRequest`s from invoices → [#197](https://github.com/GEWIS/sudosos-backend/issues/197).
- Settlement (two-way) for disabled users → [#139](https://github.com/GEWIS/sudosos-backend/issues/139). Note: `PaymentRequest` covers the *debit* direction only. The payout direction (GEWIS owes user) needs a separate primitive and is not addressed here.
- Reconciling bank-transfer payments to an open link. Surfaced as an admin-only "mark fulfilled externally" action (see below), not automated.
- Handling Stripe refunds of a paid request.

### Architecture Decisions

- **UUID as primary key** — follow the `QRAuthenticator` pattern (`BaseEntityWithoutId` + `@PrimaryColumn` on the UUID). One identifier, not two. The UUID is what is shared publicly and what internal relations reference.
- **Amount is baked in and immutable.** Normalizing against the user's balance at pay-time makes the email/SudoSOS amounts diverge whenever the user transacts between link creation and payment. Immutability is also what lets us safely cancel + reissue when an amount is wrong (see "Mistakes & cancellation" below).
- **Stripe relation is inverted.** The proposal originally placed `stripeDeposit?` on the payment request as a `OneToOne`. That breaks the retry case (user starts a payment, abandons, comes back and tries again — we need a new intent). Instead, `StripePaymentIntent` gets an optional FK to `PaymentRequest`. The set of intents for a request determines its status; "paid" is derived from the existence of a completed `StripeDeposit` with a `Transfer`.
- **Status is derived, not stored.** Like `QRAuthenticator.status`, expose a computed enum `PENDING | PAID | EXPIRED | CANCELLED`. The only stored state is `cancelledAt` and `paidAt` and `expiresAt`; the rest is read from the Stripe side and the clock.
- **Audit:** store `createdBy` so we can answer "who generated this link".
- **No coupling to `Invoice` in v1.** When #197 lands, the invoice will reference the request (`Invoice.paymentRequest?`), not the other way around. This keeps the primitive reusable for non-invoice origins (admin-issued top-ups, board deposits, etc.).
- **Always credits balance on successful payment.** A successful Stripe deposit on a `PaymentRequest` always results in a standard void→user `Transfer`. `PaymentRequest` is a top-up primitive; callers that do not want a balance credit (e.g. automagic invoices in #197) must not pre-create their own invoice-Transfer. See the second edit note below for why this decomposition is correct.

### Entity sketch

```typescript
@Entity()
export default class PaymentRequest extends BaseEntityWithoutId {
  @PrimaryColumn({ type: 'varchar', length: 36 })
  public id: string; // UUID v4, set in constructor

  @JoinColumn()
  public for: User; // User account to top up

  @ManyToOne(() => User, { nullable: false })
  @JoinColumn()
  public createdBy: User; // Audit: who issued this request

  @Column({ type: 'integer', transformer: DineroTransformer.Instance })
  public amount: Dinero; // Fixed, immutable

  @Column({ type: 'datetime' })
  public expiresAt: Date;

  @Column({ type: 'datetime', nullable: true })
  public cancelledAt: Date | null;

  @Column({ type: 'datetime', nullable: true })
  public paidAt: Date | null;

  public get status(): PaymentRequestStatus {
    if (this.paidAt !== null) return PaymentRequestStatus.PAID;
    if (this.cancelledAt !== null) return PaymentRequestStatus.CANCELLED;
    if (this.expiresAt < new Date()) return PaymentRequestStatus.EXPIRED;
    return PaymentRequestStatus.PENDING;
  }
}
```

And on the Stripe side:

```typescript
// StripePaymentIntent gains an optional FK
@ManyToOne(() => PaymentRequest, { nullable: true, onDelete: 'SET NULL' })
@JoinColumn()
public paymentRequest?: PaymentRequest | null;
```

### Service: `PaymentRequestService`

Extends `WithManager` (mandatory — `markPaid` must participate in the Stripe webhook's existing transaction block).

**Methods**:
- `createPaymentRequest({ for, createdBy, amount, expiresInDays?, description? })` — new request, UUID v4, default expiry from config.
- `getPaymentRequest(id)` / `getPaymentRequests(filters, pagination)` — lookup + paginated list.
- `validatePayable(req)` — throws `PaymentRequestError` for expired/cancelled/paid.
- `startPayment(req)` — calls `StripeService.createStripePaymentIntent(req.for, req.amount, req)`. Safe to call repeatedly; each call mints a fresh intent. **Always results in a balance-credit Transfer on success** — do not use this if you do not want that.
- `cancelPaymentRequest(req, by)` — sets `cancelledAt`; rejects if already paid.
- `markFulfilledExternally(req, by, reason)` — admin escape hatch for bank-transfer cases. Creates the `Transfer` manually and sets `paidAt`.
- `markPaid(req, transfer)` (internal; called from Stripe webhook hook) — idempotent.

### Controller split

**Authenticated controller** (`PaymentRequestController`, mounted after token middleware):
- `POST /payment-requests` — create. Policy: `roleManager.can(..., 'create', 'all', 'PaymentRequest', ['*'])`.
- `GET /payment-requests` — paginated list with filters (`forId`, `createdById`, `status`, date range). `own`/`all` visibility via a `getRelation` helper.
- `GET /payment-requests/:id` — single request.
- `DELETE /payment-requests/:id` — cancel.
- `POST /payment-requests/:id/mark-fulfilled` — Super admin only. Body must include a reason.

**Public controller** (`PaymentRequestPublicController`, mounted before `setupAuthentication`, following the `StripeWebhookController` pattern):
- `GET /payment-requests-public/:id` — returns `PublicPaymentRequestResponse` (lean: recipient display name, amount, status, expiration — no email, balance, user type, or pricing group).
- `POST /payment-requests-public/:id/start` — validates + creates Stripe intent. Returns `{ clientSecret, publishableKey }`.

Rate-limiting of the public `start` endpoint is **not** included here (no rate-limiting middleware exists in the codebase today). To be filed as a follow-up issue covering all public endpoints.

### Validation rules on `for`

- `UserType.INVOICE` — **allowed** (needed for #197's automagic invoice flow).
- Inactive users — **allowed** (there may be legitimate reasons to top up an inactive account before reactivation).
- `UserType.POINT_OF_SALE` — **rejected** (not a human payer).
- Soft-deleted users — **rejected**.

### StripeService changes

- Extend `createStripePaymentIntent(user, amount, paymentRequest?)` to persist the FK on `StripePaymentIntent` and add `paymentRequestId: req.id` to Stripe metadata (for webhook correlation even if our DB lags).
- In `createNewPaymentIntentStatus`, extend the inner `findOne` relations with `paymentRequest: true`. In the `SUCCEEDED` branch, after the transfer is saved, if `paymentIntent.paymentRequest && !paymentIntent.paymentRequest.paidAt`, call `new PaymentRequestService(this.manager).markPaid(...)`. Inherits the status-level idempotency for free.

## Mistakes & cancellation

- **Wrong amount / wrong user** → cancel the request, create a new one. Do not mutate amounts.
- **Paid but shouldn't have been** → Stripe refund flow, tracked separately (out of scope here).
- **User paid by bank transfer instead of iDeal** → admin uses `POST /payment-requests/:id/mark-fulfilled` after reconciliation. Frontend should make iDeal the obvious default and put "view as invoice" / bank-transfer details in secondary UI.

## Security notes

- UUIDs are intended to be shareable. Anyone with the link can pay, but the public endpoint must leak only display name + amount + expiration + status.
- All state-changing endpoints require authentication and appropriate RBAC roles. Suggested role wiring:
  - **Super admin**: full `admin` CRUD on `PaymentRequest`.
  - **AuthorizedBuyer** (kiosk/cashier): `create + get` (so cashiers can mint links for a user on the spot).
  - **User**: `get own` + `delete own` (users can see and cancel links issued for them).

## Future: how this composes with #197 and #139

- **#197 (Automagic Invoices):** invoice generation creates a `PaymentRequest` and references it from `Invoice.paymentRequest?`. Automagic invoices do **not** pre-create their own invoice-Transfer (unlike manual invoices); the Stripe deposit is the settlement. This avoids double-crediting the balance. See the second edit note below.
- **#139 (Final "invoice" for disabled users):** negative-balance case → create a `PaymentRequest`. Positive-balance case (GEWIS owes user) → **not** covered by this primitive; needs a separate payout mechanism.

### Other information

[Based on this comment](https://github.com/GEWIS/sudosos-backend/issues/197#issuecomment-2258780616)

---

**edit (1):** rewrote the proposal after a design review with @Yoronex and @RubenLWF. The primitive is now called `PaymentRequest` (the "payment link" is just how it's delivered); UUID became the primary key (QR-authenticator pattern); the `StripeDeposit`/`StripePaymentIntent` relation is inverted so retries work; amount stays immutable (Yoronex's email-mismatch argument won); status is derived rather than stored; `createdBy` added for audit; a rate limit and an admin "mark fulfilled externally" escape hatch were added; and the `Invoice` coupling was moved out to #197. Scope is now explicitly the primitive only — #197 and #139 compose on top.

**edit (2):** clarified the balance-credit semantics after a follow-up design question about invoice accounts. `PaymentRequest` is a pure top-up primitive — a successful payment **always** creates a standard void→user `Transfer`, regardless of the recipient's `UserType`. The "don't double-credit when an automagic invoice is settled" concern is not solved by special-casing this primitive; it is solved in #197 by having automagic invoices **not** pre-create their own invoice-Transfer at generation time. The Stripe deposit becomes the settlement event. This keeps the primitive context-free and pushes the bookkeeping decision to the one place that has context to make it. Added validation rules on the `for` user (invoice type and inactive users allowed; POS-type and soft-deleted rejected). Added a test guaranteeing balance credit on invoice-typed recipients. Corresponding note posted on #197.

JustSamuel:
# Description

Implements the `PaymentRequest` primitive from #639: an unauthenticated, fixed-amount, expiring request to top up a specific user's SudoSOS balance. End-users see it as a shareable "payment link"; in the domain it is a request entity with derived status.

This PR is built up commit-by-commit so the story is reviewable. The full design is captured in #639 (with two `edit:` notes documenting the consensus reached in the issue thread). High-level shape:

- **Entity**: UUID PK (QR-authenticator pattern), immutable `amount` (Dinero), `expiresAt`, `paidAt`/`cancelledAt`/`cancelledBy` for state, `for`/`createdBy` for audit. Derived `status` getter (`PENDING | PAID | EXPIRED | CANCELLED`).
- **Stripe relation inverted**: `StripePaymentIntent.paymentRequest?` — supports retries (one request, many intents). "Paid" derived from a successful intent's existing `StripeDeposit.transfer`.
- **Always credits balance on success.** Pure top-up primitive. Automagic invoices (#197) compose by *not* pre-creating their own invoice-Transfer; the Stripe deposit becomes the settlement event.
- **Public + authenticated controllers split.** Public lookup/start endpoints mounted before token middleware (StripeWebhookController pattern). Admin operations (create/list/cancel/mark-fulfilled-externally) behind RBAC.

## Phase tracker (commits)

- [x] **Phase 1**: Entity + migrations + DB registration.
- [x] **Phase 2**: `PaymentRequestService` + Stripe webhook hook.
- [x] **Phase 3**: Authenticated controller + public controller + DTOs.
- [x] **Phase 4**: RBAC default-roles wiring.
- [x] **Phase 5**: Seeder + service tests + controller tests.
- [x] **Phase 6**: Documentation (entity prose + general docs).
- [x] **Phase 7**: Final lint + tsc + full test suite (CI gate).

## Related

Closes #639. See cross-reference comments on #197 and #139.

## Types of changes

- New feature

---

## ✅ PR Checklist

- [x] **Test Coverage**: PaymentRequestService + PaymentRequestController + PaymentRequestPublicController covered.
- [x] **Documentation**: entity module prose, Core Concepts, and Key Workflows updated.
- [x] **Database Changes**: two TypeORM migrations included (create `payment_request`, alter `stripe_payment_intent`).

### DIFF
--- a/docs/content/general/1-core-concepts.md
+++ b/docs/content/general/1-core-concepts.md
@@ -91,6 +91,18 @@ An **invoice** is best understood as a **top-up with a paper trail**.
 
 **VAT note**: the invoice transfer itself is **not** VAT-applicable. VAT is handled by the underlying transactions and shows up in the relevant seller payouts.
 
+## Payment requests (shareable payment links)
+
+A **payment request** is a fixed-amount, time-bounded invitation to top up a specific user's balance. End-users recognise it as a "payment link": an email with an amount and a URL that anyone (including the user themselves from a different device) can open to pay via Stripe.
+
+- The request is created in `PENDING` state and carries an `expiresAt` plus an immutable `amount`.
+- A user can retry: one request can spawn many `StripePaymentIntent` rows. The request becomes `PAID` when one of those intents succeeds.
+- A successful payment **always credits the recipient's balance** via the standard Stripe deposit transfer (void → user). It is a pure top-up primitive.
+- Cancellation is explicit and audit-tracked (`cancelledBy`, `cancelledAt`). To "change the amount", cancel and create a new one.
+- An admin escape hatch marks a request fulfilled out-of-band (e.g. bank transfer) — this creates the credit transfer manually.
+
+Payment requests and **invoices** overlap conceptually (both top up a balance) but differ in who owns settlement: invoices pre-create their own credit transfer, payment requests delegate to the Stripe deposit. A future "invoice via payment link" flow will compose them by having the invoice skip pre-creating a transfer when a payment request is attached.
+
 ## Seller payouts (settling sellers and recording VAT)
 
 A **seller payout** is a payout transfer for a seller over a time range:
--- a/docs/content/general/2-key-workflows.md
+++ b/docs/content/general/2-key-workflows.md
@@ -87,6 +87,40 @@ flowchart TD
 - event routing: only accept events intended for this service
 - idempotency: the same Stripe event may be delivered multiple times
 
+## Payment request (shareable payment link)
+
+**Trigger + actor**
+- Treasurer (or a user for themself) creates a fixed-amount top-up request with an expiry. The recipient opens a share link (no login), pays via Stripe, and the money is credited to the linked user's balance.
+
+**API surface**
+- `POST /payment-requests` (create; admin can create for anyone, regular users only for themselves)
+- `GET /payment-requests` and `GET /payment-requests/:id` (read; admin sees all, user sees own)
+- `POST /payment-requests/:id/cancel` (cancel a PENDING request)
+- `POST /payment-requests/:id/start` (authenticated: start a Stripe session)
+- `POST /payment-requests/:id/mark-fulfilled` (admin escape hatch for out-of-band payment)
+- `GET /payment-requests-public/:id` (unauthenticated: fetch a trimmed view via share link)
+- `POST /payment-requests-public/:id/start` (unauthenticated: start a Stripe session via share link)
+
+**Entities touched**
+- `PaymentRequest` (UUID pk, immutable amount, derived status)
+- `StripePaymentIntent.paymentRequest?` — one request can have many intents (retries)
+- on success: the standard Stripe deposit `Transfer` (void → user) already created by the deposit flow also settles the payment request
+
+**Lifecycle (derived status)**
+- `PENDING` → `PAID` when a linked Stripe intent succeeds.
+- `PENDING` → `CANCELLED` when an authorized user cancels.
+- `PENDING` → `EXPIRED` when `expiresAt` passes without payment.
+- Status is derived from `paidAt` / `cancelledAt` / `expiresAt`; precedence is `PAID` > `CANCELLED` > `EXPIRED` > `PENDING`.
+
+**Critical checks**
+- amount is fixed at creation — a corrected amount means cancelling and creating a fresh request
+- `startPayment` rejects non-`PENDING` requests
+- the Stripe webhook flow idempotently calls `markPaid`: already-paid requests are left unchanged, cancelled requests refuse to flip, and expired-but-settled requests still become `PAID` (the money arrived, the clock is secondary)
+
+**Invoices vs. payment requests**
+- Invoices create their own credit transfer on behalf of the user (a "paper-trail top-up").
+- Payment requests do not pre-create a credit transfer — the Stripe deposit is the single settlement event. A future "invoice via payment link" flow will compose by leaving the settlement to the payment request.
+
 ## Payout request (withdraw balance)
 
 **Trigger + actor**
--- a/src/controller/payment-request-controller.ts
+++ b/src/controller/payment-request-controller.ts
@@ -1,0 +1,414 @@
+/**
+ *  SudoSOS back-end API service.
+ *  Copyright (C) 2026 Study association GEWIS
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as published
+ *  by the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ *  @license
+ */
+
+/**
+ * This is the module page of the payment-request-controller.
+ *
+ * Authenticated endpoints for managing {@link stripe/payment-request!PaymentRequest | PaymentRequest}
+ * rows. The unauthenticated share-link surface lives in
+ * {@link PaymentRequestPublicController}.
+ *
+ * @module stripe/payment-request
+ */
+
+import { Response } from 'express';
+import log4js, { Logger } from 'log4js';
+import Dinero from 'dinero.js';
+import BaseController, { BaseControllerOptions } from './base-controller';
+import Policy from './policy';
+import { RequestWithToken } from '../middleware/token-middleware';
+import { parseRequestPagination, toResponse } from '../helpers/pagination';
+import PaymentRequestService, {
+  IllegalPaymentRequestTransitionError,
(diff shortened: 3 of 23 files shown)


JustSamuel:
Superseded by a 3-PR stack for digestible review:

- **#895** — `Add PaymentRequest schema and migration` (entity + migration; base `develop`)
- **#896** — `Add PaymentRequest service layer` (service + Stripe hook + service tests + seeder; base #895)
- **#897** — `Add PaymentRequest HTTP API` (controllers + RBAC + controller tests + docs; base #896)

Same 7 commits, redistributed across three branches with no content changes. Each branch builds + lints + passes tests on its own. Closing this PR once the stack is reviewed; leaving open for now in case reviewers want a single-diff reference.

PR #895, opened 2026-05-01
Add PaymentRequest schema and migration

JustSamuel:
# Description

**Stack: 1 of 3** — schema layer of the [PaymentRequest primitive (#639)](https://github.com/GEWIS/sudosos-backend/issues/639). This PR ships only the entity and migration; the service and HTTP layers follow in dependent PRs.

A PaymentRequest is an unauthenticated, fixed-amount, expiring request to top up a specific user's SudoSOS balance. End-users see it as a shareable "payment link"; in the domain it is a request entity with derived status (`PENDING | PAID | EXPIRED | CANCELLED`).

## What this PR adds

- **`PaymentRequest` entity** — UUID PK (QR-authenticator pattern), immutable `amount` (Dinero), `expiresAt`, `paidAt` / `cancelledAt` / `cancelledBy` / `fulfilledBy` for state, `for` / `createdBy` for audit. Derived `status` getter computed from stored timestamp columns.
- **Migration `1777010230727-payment-request.ts`** — creates the `payment_request` table with indexes on `forId`, `createdById`, `cancelledById`, `fulfilledById`, and `expiresAt`; alters `stripe_payment_intent` to add the optional `paymentRequestId` FK so a single request can be retried across many intents.
- **Stripe relation inverted** — `StripePaymentIntent.paymentRequest?` lives on the intent side, so retries are first-class. "Paid" is derived from a successful intent's existing `StripeDeposit.transfer` (no parallel ledger).

The columns are dormant in this PR — the next stack PR adds the service that reads/writes them.

Part of #639. Continues in:
- PR #2: `Add PaymentRequest service layer` (depends on this)
- PR #3: `Add PaymentRequest HTTP API` (depends on #2)

## Types of changes

- New feature

---

## ✅ PR Checklist

- [x] **Test Coverage**: schema-only PR — entity behavior is exercised by service tests in the next stack PR. Migration up/down both run cleanly under `pnpm run schema`.
- [x] **Documentation**: TypeDoc prose on the entity module + per-field JSDoc.
- [x] **Database Changes**: migration included, tested with both SQLite (vitest) and MariaDB (CI coverage job).

## 🔗 Additional Notes

This is the first PR in a 3-way stack splitting [#890](https://github.com/GEWIS/sudosos-backend/pull/890). The original combined PR will be closed once the stack lands.

### DIFF
--- a/src/entity/payment-request/payment-request.ts
+++ b/src/entity/payment-request/payment-request.ts
@@ -1,0 +1,222 @@
+/**
+ *  SudoSOS back-end API service.
+ *  Copyright (C) 2026 Study association GEWIS
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as published
+ *  by the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ *  @license
+ */
+
+/**
+ * This is the module page of the payment-request.
+ *
+ * A `PaymentRequest` is an immutable, unauthenticated, shareable request to
+ * top up a specific user's balance with a specific amount before a specific
+ * deadline. It is the domain primitive behind what end-users perceive as a
+ * "payment link".
+ *
+ * ## Lifecycle
+ *
+ * A request is created in `PENDING` state with an `expiresAt` and a fixed
+ * `amount`. From there it transitions to exactly one of:
+ *
+ * - `PAID` — a linked {@link stripe!StripePaymentIntent | StripePaymentIntent}
+ *   succeeded and the corresponding void->user `Transfer` was created.
+ * - `CANCELLED` — explicitly cancelled by an authorized user (e.g. a wrong
+ *   amount was issued and a fresh request will be created instead).
+ * - `EXPIRED` — `expiresAt` has passed without payment.
+ *
+ * Status is **derived**, not stored — see {@link PaymentRequest.status}. The
+ * only stored state is `paidAt`, `cancelledAt`, and `expiresAt`.
+ *
+ * ## Stripe relationship (one-to-many on the inverse side)
+ *
+ * A request may have many {@link stripe!StripePaymentIntent | StripePaymentIntent}
+ * rows over its lifetime — each call to `startPayment` mints a fresh intent
+ * (the user can abandon a session and try again). The request becomes "paid"
+ * when at least one of those intents reaches `SUCCEEDED` and its
+ * `StripeDeposit.transfer` exists.
+ *
+ * ## Amount immutability
+ *
+ * The `amount` is fixed at creation time and cannot be mutated. The user-
+ * facing email/PDF mentions a specific euro figure; if SudoSOS were to
+ * recompute the figure at pay-time (against a moving balance), the email
+ * and the actual charge would diverge. To correct a wrong amount, cancel
+ * the request and create a new one — never mutate.
+ *
+ * ## Always credits balance on success
+ *
+ * A successful payment **always** results in a standard void->user `Transfer`
+ * that credits the recipient's balance. `PaymentRequest` is a top-up
+ * primitive — callers that do not want a balance credit (e.g. a future
+ * "invoice via payment link" flow) must not pre-create their own credit
+ * Transfer; the Stripe deposit is the single settlement event.
+ *
+ * ## Admin escape hatch
+ *
+ * Reality intrudes: occasionally a user pays a payment-link's invoice via
+ * bank transfer instead of iDeal. The admin endpoint
+ * `POST /payment-requests/:id/mark-fulfilled` creates the credit Transfer
+ * manually (with a reason for audit) — this credits the recipient's balance
+ * exactly the same way a Stripe payment would — and flips the request to
+ * `PAID` without a Stripe deposit existing.
+ *
+ * ## Out of scope (tracked elsewhere)
+ *
+ * - Linking back to {@link invoicing!Invoice | Invoice}
+ *   (`Invoice.paymentRequest?` lives on the invoice side).
+ * - Settlement (positive-balance payouts) for disabled users — a separate
+ *   primitive, not this one.
+ * - Stripe refund handling.
+ *
+ * @module stripe/payment-request
+ * @mergeTarget
+ */
+
+import {
+  Column, Entity, JoinColumn, ManyToOne, PrimaryColumn,
+} from 'typeorm';
+import { v4 as uuidv4 } from 'uuid';
+import { Dinero } from 'dinero.js';
+import BaseEntityWithoutId from '../base-entity-without-id';
+import User from '../user/user';
+import DineroTransformer from '../transformer/dinero-transformer';
+import { PaymentRequestStatus } from './payment-request-status';
+
+/**
+ * A shareable, fixed-amount request to top up a specific user's SudoSOS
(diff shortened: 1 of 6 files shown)


src/migrations/1777010230727-payment-request.ts
```diff
@@ -0,0 +1,206 @@
...
+          type: 'integer',
+          isNullable: true,
+        },
+        {
+          name: 'paidAt',
+          type: 'datetime',
+          isNullable: true,
+        },
+        {
+          name: 'fulfilledById',
```
RubenLWF, line 77: suggestion:

I think we should call this `paidById`, just like how we have `cancelledAt` and `cancelledById`.

src/migrations/1777010230727-payment-request.ts
```diff
@@ -0,0 +1,206 @@
...
+          type: 'integer',
+          isNullable: true,
+        },
+        {
+          name: 'description',
```
RubenLWF, line 82: question:

What do we need this field for?

src/migrations/1777010230727-payment-request.ts
```diff
@@ -0,0 +1,206 @@
...
+        },
+        {
+          name: 'updatedAt',
+          type: 'datetime',
+          default: 'current_timestamp',
+          onUpdate: 'current_timestamp',
+          isNullable: false,
+        },
+        {
+          name: 'version',
```
RubenLWF, line 101: question:

In what scenarios will a payment request get a new version? Won't it just become a new payment request?

src/entity/payment-request/payment-request.ts
```diff
@@ -0,0 +1,216 @@
...
+ * deadline. It is the domain primitive behind what end-users perceive as a
+ * "payment link".
+ *
+ * ## Lifecycle
+ *
+ * A request is created in `PENDING` state with an `expiresAt` and a fixed
+ * `amount`. From there it transitions to exactly one of:
+ *
+ * - `PAID` — a linked {@link stripe!StripePaymentIntent | StripePaymentIntent}
+ *   succeeded and the corresponding void→user `Transfer` was created.
```
RubenLWF: nitpick:

Using unicode arrows in comments really rubs me the wrong way.

src/entity/payment-request/payment-request.ts
```diff
@@ -0,0 +1,216 @@
...
+ * when at least one of those intents reaches `SUCCEEDED` and its
+ * `StripeDeposit.transfer` exists.
+ *
+ * ## Amount immutability
+ *
+ * The `amount` is fixed at creation time and cannot be mutated. The user-
+ * facing email/PDF mentions a specific euro figure; if SudoSOS were to
+ * recompute the figure at pay-time (against a moving balance), the email
+ * and the actual charge would diverge. To correct a wrong amount, cancel
+ * the request and create a new one — never mutate.
```
RubenLWF, line 57: big fan of this philosophy

src/entity/payment-request/payment-request.ts
```diff
@@ -0,0 +1,216 @@
...
+ * "invoice via payment link" flow) must not pre-create their own credit
+ * Transfer; the Stripe deposit is the single settlement event.
+ *
+ * ## Admin escape hatch
+ *
+ * Reality intrudes: occasionally a user pays a payment-link's invoice via
+ * bank transfer instead of iDeal. The admin endpoint
+ * `POST /payment-requests/:id/mark-fulfilled` creates the credit Transfer
+ * manually (with a reason for audit), flipping the request to `PAID` without
+ * a Stripe deposit existing.
```
RubenLWF: questoin:

This will also increase the balance of the user? Or just mark the payment request as `PAID`?

src/entity/payment-request/payment-request.ts
```diff
@@ -0,0 +1,216 @@
...
+  public createdBy: User;
+
+  /** Fixed, immutable amount. */
+  @Column({
+    type: 'integer',
+    transformer: DineroTransformer.Instance,
+  })
+  public amount: Dinero;
+
+  /** After this moment the request stops accepting payments. */
```
RubenLWF: question:

What status will a request get once it expires? `CANCELLED`?

src/entity/payment-request/payment-request.ts
```diff
@@ -0,0 +1,216 @@
...
+    nullable: true,
+  })
+  public description: string | null;
+
+  /**
+   * Derived status — not persisted. Order of precedence:
+   * `PAID` > `CANCELLED` > `EXPIRED` > `PENDING`.
+   */
+  public get status(): PaymentRequestStatus {
+    if (this.paidAt !== null && this.paidAt !== undefined) {
```
RubenLWF: suggestion:

Since both values are falsy:
```suggestion
    if (!!this.paidAt) {
```

src/entity/payment-request/payment-request.ts
```diff
@@ -0,0 +1,216 @@
...
+      return PaymentRequestStatus.PAID;
+    }
+    if (this.cancelledAt !== null && this.cancelledAt !== undefined) {
```
RubenLWF: suggestion:

Since both values are falsy:
```suggestion
    if (!!this.cancelledAt) {
```

RubenLWF (changes requested):

src/migrations/1777010230727-payment-request.ts
JustSamuel: Renaming to paidById would imply it's set whenever a request is paid, which would be misleading. It is set only by an admin as an escape hatch (markFulfilledExternally)

JustSamuel: Naming is intentional. This column is set **only** on the admin escape hatch path (`markFulfilledExternally`), never on the Stripe-paid path where the actor is implicitly the user themselves. Renaming to `paidById` would suggest it gets populated whenever a request is paid, which is misleading because Stripe-settled requests always leave this column NULL. Keeping `fulfilledBy` to make the audit-only semantic explicit.

src/migrations/1777010230727-payment-request.ts
JustSamuel: End-user-facing context label shown on the share-link landing page, so the recipient knows what they are paying for (e.g. `"Drinks at GEWIS FSE naborrel 2026-04-12"` or `"Invoice #123"`). The dashboard `Create payment request` form lets the issuing admin fill it in. Already documented on the entity field and in the `@typedef property` tag.

src/migrations/1777010230727-payment-request.ts
JustSamuel: TypeORM optimistic-locking column. Auto-increments on every `save()` call, not on logical state changes. Every entity in the codebase carries it via `BaseEntity` / `BaseEntityWithoutId`, so it is not specific to PaymentRequest. Defined here in the migration only because we register the entity by extending `BaseEntityWithoutId` (UUID primary key) rather than `BaseEntity`.

src/entity/payment-request/payment-request.ts
JustSamuel: Replaced both unicode arrows with ASCII `->` in the module docstring.

src/entity/payment-request/payment-request.ts
JustSamuel: It also credits the balance. Updated the `Admin escape hatch` paragraph to spell out that the manual Transfer credits the recipient the same way a Stripe payment would, and only then flips the request to PAID. A successful payment-request settlement always results in a balance credit; this path is no exception.

src/entity/payment-request/payment-request.ts
JustSamuel: `EXPIRED`. There is a separate enum value for it, never `CANCELLED`. Expanded the `expiresAt` field doc at point of definition with a cross-link to `PaymentRequest.status` for the precedence rules (`PAID` > `CANCELLED` > `EXPIRED` > `PENDING`).

src/entity/payment-request/payment-request.ts
JustSamuel: Applied the suggestion.

src/entity/payment-request/payment-request.ts
JustSamuel: Applied the suggestion.

RubenLWF (approved):

PR #896, opened 2026-05-01
Add PaymentRequest service layer

JustSamuel:
# Description

**Stack: 2 of 3** — business logic for the [PaymentRequest primitive (#639)](https://github.com/GEWIS/sudosos-backend/issues/639). Builds on PR #895 (schema). The HTTP surface follows in PR #897.

This PR is the full **service layer**: the state machine, response DTOs the service emits, the Stripe settlement integration, and the test seeder + service test suite. **No controller code** ships in this PR — the schema and the service-side response types are everything needed for the next PR's controllers to compile against.

### Service (`src/service/payment-request-service.ts`)
- **Creation & validation** — beneficiary eligibility (rejects POS users, soft-deleted users), strictly-future `expiresAt`, strictly-positive `amount`, 255-char description limit matching `varchar(255)`.
- **Lookup** — `getPaymentRequest(id)` (full audit relations) + `getPublicPaymentRequest(id)` (lightweight, only `for`).
- **Listing** — `getPaymentRequests(filters, pagination)` runs derived-status filtering as SQL predicates on `paidAt`/`cancelledAt`/`expiresAt` so pagination happens in the DB.
- **State transitions** — `cancelPaymentRequest`, `markFulfilledExternally` (admin escape hatch with persisted `fulfilledBy` actor + audit description respecting `Transfer.description` length), `markPaid` (idempotent), `markPaidFromStripeIntent` (webhook hot-path with no relation joins).
- **Mappers** — `asBasePaymentRequestResponse`, `asPublicPaymentRequestResponse`, plus `parseGetPaymentRequestsFilters` query-string parser.

### Response types (`src/controller/response/`)
- `payment-request-response.ts` — `BasePaymentRequestResponse`, `PublicPaymentRequestResponse`, `PaymentRequestStartResponse`.
- `base-response-without-id.ts` — new shared base extracted so non-integer-keyed responses (UUID-keyed `BasePaymentRequestResponse`) can share timestamp/version fields with `BaseResponse`.
- `base-response.ts` — refactored to extend `BaseResponseWithoutId`.

### Stripe webhook hook (`src/service/stripe-service.ts`)
- `createStripePaymentIntent` accepts an optional `paymentRequest` and stores the back-reference.
- `createNewPaymentIntentStatus` flips the linked request to PAID **best-effort** on `SUCCEEDED` — wrapped in try/catch so a state-machine conflict cannot roll back the credit Transfer that has already settled with Stripe.

### Tests + seeder
[...]

PR #897, opened 2026-05-01
Add PaymentRequest HTTP API

### LINKED ISSUE #639: [Feature]: PaymentRequest entity

JustSamuel:
# Description

**Stack: 3 of 3** — HTTP surface for the [PaymentRequest primitive (#639)](https://github.com/GEWIS/sudosos-backend/issues/639). Builds on PR #896 (service), which builds on PR #895 (schema).

Pure controller PR: this branch ships the authenticated controller, the unauthenticated share-link controller, request DTOs, RBAC wiring, controller tests, the user-facing documentation, and the existing `UserController` extension that lists a user's payment requests. **No service code** lives in this PR — every method called on the service is already in #896.

### Authenticated controller (`/payment-requests`)
- `POST /` — create a request (admin or self).
- `GET /` — list with filters + pagination; users with `get:own` only see their own (`forId` is server-forced).
- `GET /:id` — fetch single (cached on the request object via `loadPaymentRequest` so policy + handler don't double-query).
- `POST /:id/cancel` — PENDING → CANCELLED.
- `POST /:id/start` — bootstrap a Stripe payment session for one's own request.
- `POST /:id/mark-fulfilled` — admin escape hatch when a user paid out-of-band.

### Public controller (`/payment-requests-public/:id`)
- Mounted **before** `setupAuthentication` in `src/index.ts`, mirroring `StripeWebhookController`. Every endpoint policy is `async () => true` — the share link IS the credential.
- `GET /:id` returns a trimmed `PublicPaymentRequestResponse` (omits `createdBy`, `cancelledBy`, `fulfilledBy`); uses the lightweight `getPublicPaymentRequest` from #896 to avoid loading audit joins.
- `POST /:id/start` boots a Stripe session without authentication.

### `UserController` extension
- `GET /users/:id/payment-requests` — lists payment requests for a specific user, gated by `get:own` (caller's own id) or `get:all`.

### Request DTOs (`src/controller/request/payment-request-request.ts`)
- `CreatePaymentRequestRequest`, `MarkFulfilledExternallyRequest`.

### RBAC (`src/rbac/default-roles.ts`)
- `Local Admin` — full `create:all`, `update:all`, `get:all`, `delete:all` on PaymentRequest.
- `User` — `get:own` (so the dashboard can list a user's own requests).

### Tests + docs
- `test/unit/controller/payment-request-controller.ts` + `payment-request-public-controller.ts` — 36 tests covering both surfaces, including get-own scoping (caller cannot spoof `forId`) and policy gating.
- `docs/content/general/1-core-concepts.md` and `2-key-workflows.md` — describe the PaymentRequest primitive in the same style as the rest of the general docs.

Closes #639 once the full stack is merged. Stacked on top of PR #896.

## Types of changes

- New feature

---

## ✅ PR Checklist

- [x] **Test Coverage**: 36 new controller tests covering both auth and public surfaces. Policy resolution path is covered.
- [x] **Documentation**: TypeDoc prose on both controllers; `docs/content/general/` updated with the new primitive in Core Concepts and Key Workflows.
- [x] **Database Changes**: none in this PR (schema lives in #895).

## 🔗 Additional Notes

Stack 3 of 3. Review base is `feat/payment-request-service`. Will be retargeted to `develop` after #895 and #896 land in order.

### DIFF
--- a/src/controller/payment-request-controller.ts
+++ b/src/controller/payment-request-controller.ts
@@ -1,0 +1,414 @@
+/**
+ *  SudoSOS back-end API service.
+ *  Copyright (C) 2026 Study association GEWIS
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as published
+ *  by the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ *  @license
+ */
+
+/**
+ * This is the module page of the payment-request-controller.
+ *
+ * Authenticated endpoints for managing {@link stripe/payment-request!PaymentRequest | PaymentRequest}
+ * rows. The unauthenticated share-link surface lives in
+ * {@link PaymentRequestPublicController}.
+ *
+ * @module stripe/payment-request
+ */
+
+import { Response } from 'express';
+import log4js, { Logger } from 'log4js';
+import Dinero from 'dinero.js';
+import BaseController, { BaseControllerOptions } from './base-controller';
+import Policy from './policy';
+import { RequestWithToken } from '../middleware/token-middleware';
+import { parseRequestPagination, toResponse } from '../helpers/pagination';
+import PaymentRequestService, {
+  IllegalPaymentRequestTransitionError,
+  InvalidPaymentRequestBeneficiaryError,
+  parseGetPaymentRequestsFilters,
+} from '../service/payment-request-service';
+import PaymentRequestCheckoutService from '../service/payment-request-checkout-service';
+import PaymentRequest from '../entity/payment-request/payment-request';
+import User from '../entity/user/user';
+import { PaymentRequestStartResponse } from './response/payment-request-response';
+import {
+  CreatePaymentRequestRequest,
+  MarkFulfilledExternallyRequest,
+} from './request/payment-request-request';
+
+export default class PaymentRequestController extends BaseController {
+  private logger: Logger = log4js.getLogger('PaymentRequestController');
+
+  public constructor(options: BaseControllerOptions) {
+    super(options);
+    this.configureLogger(this.logger);
+  }
+
+  /**
+   * @inheritDoc
+   */
+  public getPolicy(): Policy {
+    return {
+      '/': {
+        GET: {
+          policy: async (req) => this.roleManager.can(
+            req.token.roles, 'get', 'all', 'PaymentRequest', ['*'],
+          ),
+          handler: this.returnAllPaymentRequests.bind(this),
+        },
+        POST: {
+          policy: async (req) => this.roleManager.can(
+            req.token.roles, 'create',
+            await PaymentRequestController.getRelation(req),
+            'PaymentRequest', ['*'],
+          ),
+          handler: this.createPaymentRequest.bind(this),
+          body: { modelName: 'CreatePaymentRequestRequest' },
+        },
+      },
+      '/:id': {
+        GET: {
+          policy: async (req) => this.roleManager.can(
+            req.token.roles, 'get',
+            await PaymentRequestController.getRelation(req),
+            'PaymentRequest', ['*'],
+          ),
+          handler: this.returnSinglePaymentRequest.bind(this),
+        },
+      },
+      '/:id/cancel': {
+        POST: {
+          policy: async (req) => this.roleManager.can(
+            req.token.roles, 'update',
+            await PaymentRequestController.getRelation(req),
+            'PaymentRequest', ['*'],
+          ),
+          handler: this.cancelPaymentRequest.bind(this),
+        },
+      },
+      '/:id/start': {
+        POST: {
+          policy: async (req) => this.roleManager.can(
+            req.token.roles, 'update',
+            await PaymentRequestController.getRelation(req),
+            'PaymentRequest', ['*'],
+          ),
+          handler: this.startPaymentAuthenticated.bind(this),
+        },
+      },
+      '/:id/mark-fulfilled': {
+        POST: {
+          policy: async (req) => this.roleManager.can(
+            req.token.roles, 'update', 'all', 'PaymentRequest', ['*'],
+          ),
+          handler: this.markFulfilledExternally.bind(this),
+          body: { modelName: 'MarkFulfilledExternallyRequest' },
+        },
+      },
+    };
+  }
+
+  /**
+   * Load the PaymentRequest for the current `/:id` request, caching on `req`
+   * so the policy + handler together only hit the DB once.
+   *
+   * The full relation set (`for`, `createdBy`, `cancelledBy`, `fulfilledBy`)
+   * is loaded so that handlers can serialize the response directly from
+   * the cached entity without re-querying.
+   *
+   * Returns `null` when there is no `:id` param or the request doesn't exist.
+   */
+  public static async loadPaymentRequest(
+    req: RequestWithToken & { paymentRequest?: PaymentRequest | null },
+  ): Promise<PaymentRequest | null> {
+    if (req.paymentRequest !== undefined) {
+      return req.paymentRequest;
+    }
+    const { id } = req.params;
+    if (!id) {
+      req.paymentRequest = null;
+      return null;
+    }
+    req.paymentRequest = await new PaymentRequestService().getPaymentRequest(id);
+    return req.paymentRequest;
+  }
+
+  /**
+   * Determine "own" vs "all" for RBAC. Creation takes the `forId` from the
+   * body; state transitions look up the request by id and compare against
+   * the caller. The lookup is cached on `req` via
+   * {@link loadPaymentRequest} so the subsequent handler does not re-query.
+   */
+  public static async getRelation(req: RequestWithToken): Promise<string> {
+    const body = req.body as Partial<CreatePaymentRequestRequest> | undefined;
+    if (body && body.forId != null) {
+      return body.forId === req.token.user.id ? 'own' : 'all';
+    }
+
+    const request = await PaymentRequestController.loadPaymentRequest(
+      req as RequestWithToken & { paymentRequest?: PaymentRequest | null },
+    );
+    return (request != null && request.for.id === req.token.user.id) ? 'own' : 'all';
+  }
+
+  /**
+   * GET /payment-requests
+   * @summary List PaymentRequests (paginated, with filtering by beneficiary, creator, and status)
+   * @operationId getAllPaymentRequests
+   * @tags paymentRequests - Operations of the payment-request controller
+   * @security JWT
+   * @param {integer} forId.query - Filter by beneficiary user id.
+   * @param {integer} createdById.query - Filter by creator user id.
+   * @param {string} status.query - enum:PENDING,PAID,EXPIRED,CANCELLED - Comma-separated list of derived statuses.
+   * @param {string} fromDate.query - Filter requests created on or after this ISO date (inclusive).
+   * @param {string} tillDate.query - Filter requests created strictly before this ISO date (exclusive).
+   * @param {integer} take.query - How many rows the endpoint should return
+   * @param {integer} skip.query - How many rows to skip (for pagination)
+   * @return {PaginatedBasePaymentRequestResponse} 200 - All existing payment requests
+   * @return {string} 400 - Validation error
+   * @return {string} 500 - Internal server error
+   */
+  public async returnAllPaymentRequests(req: RequestWithToken, res: Response): Promise<void> {
+    this.logger.trace('Get all payment requests by user', req.token.user);
+
+    let filters;
+    let pagination;
+    try {
+      filters = parseGetPaymentRequestsFilters(req);
+      pagination = parseRequestPagination(req);
+    } catch (e) {
+      res.status(400).send(e instanceof Error ? e.message : String(e));
+      return;
+    }
+
+    try {
+      const service = new PaymentRequestService();
+      const [rows, count] = await service.getPaymentRequests(filters, pagination);
+      const records = rows.map((r) => PaymentRequestService.asBasePaymentRequestResponse(r));
+      res.status(200).json(toResponse(records, count, pagination));
+    } catch (e) {
+      this.logger.error('Could not list payment requests:', e);
+      res.status(500).send('Internal server error.');
+    }
+  }
+
+  /**
+   * GET /payment-requests/{id}
+   * @summary Fetch a single PaymentRequest by id.
+   * @operationId getSinglePaymentRequest
+   * @tags paymentRequests - Operations of the payment-request controller
+   * @param {string} id.path.required - UUID v4 of the payment request.
+   * @security JWT
+   * @return {BasePaymentRequestResponse} 200 - Single payment request
+   * @return {string} 404 - Unknown id
+   * @return {string} 500 - Internal server error
+   */
+  public async returnSinglePaymentRequest(req: RequestWithToken, res: Response): Promise<void> {
+    this.logger.trace('Get single payment request by user', req.token.user, 'id', req.params.id);
+
+    try {
+      const request = await PaymentRequestController.loadPaymentRequest(
+        req as RequestWithToken & { paymentRequest?: PaymentRequest | null },
+      );
+      if (!request) {
+        res.status(404).send();
+        return;
+      }
+      res.status(200).json(PaymentRequestService.asBasePaymentRequestResponse(request));
+    } catch (e) {
+      this.logger.error('Could not get payment request:', e);
+      res.status(500).send('Internal server error.');
+    }
+  }
+
+  /**
+   * POST /payment-requests
+   * @summary Create a new PaymentRequest.
+   * @operationId createPaymentRequest
+   * @tags paymentRequests - Operations of the payment-request controller
+   * @param {CreatePaymentRequestRequest} request.body.required - The request to create
+   * @security JWT
+   * @return {BasePaymentRequestResponse} 200 - The created payment request
+   * @return {string} 400 - Validation error
+   * @return {string} 404 - Beneficiary user not found
+   * @return {string} 500 - Internal server error
+   */
+  public async createPaymentRequest(req: RequestWithToken, res: Response): Promise<void> {
+    this.logger.trace('Create payment request by user', req.token.user, 'body', req.body);
+    const body = req.body as CreatePaymentRequestRequest;
+
+    const forUser = await User.findOne({ where: { id: body.forId, deleted: false } });
+    if (!forUser) {
+      res.status(404).send('Beneficiary user not found.');
+      return;
+    }
+
+    let expiresAt: Date;
+    try {
+      expiresAt = new Date(body.expiresAt);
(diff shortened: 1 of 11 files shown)


src/controller/payment-request-controller.ts
```diff
@@ -0,0 +1,414 @@
...
+  /**
+   * GET /payment-requests/{id}
+   * @summary Fetch a single PaymentRequest by id.
+   * @operationId getSinglePaymentRequest
+   * @tags paymentRequests - Operations of the payment-request controller
+   * @param {string} id.path.required - UUID v4 of the payment request.
+   * @security JWT
+   * @return {BasePaymentRequestResponse} 200 - Single payment request
+   * @return {string} 404 - Unknown id
+   */
```
RubenLWF, line 218: suggestion:

Please include the 500 return status as well

src/controller/payment-request-controller.ts
```diff
@@ -0,0 +1,414 @@
...
+  public getPolicy(): Policy {
+    return {
+      '/': {
+        GET: {
+          policy: async (req) => this.roleManager.can(
+            req.token.roles, 'get', 'all', 'PaymentRequest', ['*'],
+          ),
+          handler: this.returnAllPaymentRequests.bind(this),
+        },
```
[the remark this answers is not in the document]
RubenLWF: This is handled in the payment-request-public-controller

src/controller/payment-request-controller.ts
```diff
@@ -0,0 +1,414 @@
...
+   * POST /payment-requests
+   * @summary Create a new PaymentRequest.
+   * @operationId createPaymentRequest
+   * @tags paymentRequests - Operations of the payment-request controller
+   * @param {CreatePaymentRequestRequest} request.body.required - The request to create
+   * @security JWT
+   * @return {BasePaymentRequestResponse} 200 - The created payment request
+   * @return {string} 400 - Validation error
+   * @return {string} 404 - Beneficiary user not found
+   */
```
RubenLWF, line 248: 500

src/controller/payment-request-controller.ts
```diff
@@ -0,0 +1,414 @@
...
+   * POST /payment-requests/{id}/cancel
+   * @summary Cancel a PENDING PaymentRequest.
+   * @operationId cancelPaymentRequest
+   * @tags paymentRequests - Operations of the payment-request controller
+   * @param {string} id.path.required - UUID v4 of the payment request.
+   * @security JWT
+   * @return {BasePaymentRequestResponse} 200 - The cancelled payment request
+   * @return {string} 404 - Unknown id
+   * @return {string} 409 - Request is not in PENDING state
+   */
```
RubenLWF, line 299: 500

src/controller/payment-request-controller.ts
```diff
@@ -0,0 +1,414 @@
...
+      const request = await PaymentRequestController.loadPaymentRequest(
+        req as RequestWithToken & { paymentRequest?: PaymentRequest | null },
+      );
+      if (!request) {
+        res.status(404).send();
+        return;
+      }
+      const service = new PaymentRequestService();
+      const cancelled = await service.cancelPaymentRequest(request, req.token.user);
+      res.status(200).json(PaymentRequestService.asBasePaymentRequestResponse(cancelled));
```
RubenLWF, line 313: question:

shouldn't this be a 204?

src/controller/payment-request-controller.ts
```diff
@@ -0,0 +1,414 @@
...
+   * POST /payment-requests/{id}/start
+   * @summary Start a Stripe payment session for the given PaymentRequest while authenticated.
+   * @operationId startPaymentRequestAuthenticated
+   * @tags paymentRequests - Operations of the payment-request controller
+   * @param {string} id.path.required - UUID v4 of the payment request.
+   * @security JWT
+   * @return {PaymentRequestStartResponse} 200 - Stripe client secret
+   * @return {string} 404 - Unknown id
+   * @return {string} 409 - Request is not in PENDING state
+   */
```
RubenLWF, line 336: 500 and 400

src/controller/payment-request-controller.ts
```diff
@@ -0,0 +1,414 @@
...
+   *   Creates a void->user credit Transfer manually.
+   * @operationId markPaymentRequestFulfilledExternally
+   * @tags paymentRequests - Operations of the payment-request controller
+   * @param {string} id.path.required - UUID v4 of the payment request.
+   * @param {MarkFulfilledExternallyRequest} request.body.required - The audit reason
+   * @security JWT
+   * @return {BasePaymentRequestResponse} 200 - The marked-paid payment request
+   * @return {string} 404 - Unknown id
+   * @return {string} 409 - Request is not in PENDING state
+   */
```
RubenLWF, line 384: 400 and 500

src/rbac/default-roles.ts
```diff
@@ -97,6 +97,14 @@ export default class DefaultRoles {
         TermsOfService: {
           get: { own: star },
         },
+        PaymentRequest: {
+          // Any regular user can list/fetch their own PaymentRequests and
+          // create new ones for themselves. Cancelling/starting is `update`
+          // on the same relation.
+          get: { own: star },
```
RubenLWF: question:

Can you stop mansplaining?

src/index.ts
```diff
@@ -251,6 +253,10 @@ export default async function createApp(): Promise<Application> {
   };
   if (config.stripe.enabled) {
     application.app.use('/v1/stripe', new StripeWebhookController(options).getRouter());
+    // Public share-link surface for PaymentRequest. Mounted BEFORE
+    // setupAuthentication so the `async () => true` policies are reachable
+    // without a JWT.
+    application.app.use('/v1/payment-requests-public', new PaymentRequestPublicController(options).getRouter());
```
RubenLWF: question:

can you stop mansplaining

src/controller/payment-request-public-controller.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   * GET /payment-requests-public/{id}
+   * @summary Fetch a PaymentRequest via the public share link. Returns a
+   *   trimmed response that omits internal audit fields.
+   * @operationId getPublicPaymentRequest
+   * @tags paymentRequestsPublic - Unauthenticated share-link surface
+   * @param {string} id.path.required - UUID v4 of the payment request.
+   * @return {PublicPaymentRequestResponse} 200 - Single payment request (trimmed shape)
+   * @return {string} 404 - Unknown id
```
RubenLWF, line 88: 500

src/controller/payment-request-public-controller.ts
```diff
@@ -0,0 +1,147 @@
...
+   * POST /payment-requests-public/{id}/start
+   * @summary Start a Stripe payment session for the given PaymentRequest
+   *   without authentication — the share link IS the credential.
+   * @operationId startPaymentRequestPublic
+   * @tags paymentRequestsPublic - Unauthenticated share-link surface
+   * @param {string} id.path.required - UUID v4 of the payment request.
+   * @return {PaymentRequestStartResponse} 200 - Stripe client secret
+   * @return {string} 404 - Unknown id
+   * @return {string} 409 - Request is not in PENDING state
+   */
```
RubenLWF, line 120: 400 and 500

RubenLWF (changes requested):

src/controller/payment-request-controller.ts
JustSamuel: done

src/controller/payment-request-controller.ts
JustSamuel: done

src/controller/payment-request-controller.ts
JustSamuel: done

src/controller/payment-request-controller.ts
JustSamuel: Kept 200 with the updated entity, same as mark-fulfilled -- saves a refetch. Happy to switch if you'd rather 204.

src/controller/payment-request-controller.ts
JustSamuel: Added 500. Cancel has no 400 path though -- only 404/409/500.

src/controller/payment-request-controller.ts
JustSamuel: done

src/rbac/default-roles.ts
JustSamuel: lol fair, gone

src/index.ts
JustSamuel: gone

src/controller/payment-request-public-controller.ts
JustSamuel: done

src/controller/payment-request-public-controller.ts
JustSamuel: done

src/controller/payment-request-controller.ts
RubenLWF: you return a 400 on line 362

RubenLWF (changes requested):

src/controller/payment-request-controller.ts
JustSamuel: Good catch. That 400 is the InvalidPaymentRequestBeneficiaryError path on /start, and it's already documented on line 332. My earlier reply mixed this thread up with /cancel, which is the one without a 400 path.

RubenLWF (approved):

PR #931, opened 2026-05-21
Migrate Invoice PDF to pdf-compiler

JustSamuel:
# Description

First of seven PDF migrations under #926. Replaces the LaTeX invoice service with an HTML one that goes through pdf-compiler instead of pdf-generator-client, and lands the F6c design as the new invoice layout.

## Visual change

The PDF now renders as the two-page F6c design: a Factuur cover with the recipient, totals headline, BTW spec table and BAC letterhead, plus a line items page with one row per sub_transaction_row (ascending by row id). Cached invoices regenerate on next access because the new parameter shape produces different hashes, so getOrCreatePdf falls through to createPdf for any invoice with a stale PDF on disk.

The "Attn. <name>" line ("t.a.v." in Dutch) renders between the addressee and the street when invoice.attention is set. The dev seed leaves attention empty so the smoke render looks identical to before, but invoices created via the API can have it set and will now show it.

## Spillover for long invoices

Invoices with many distinct products spill onto multiple line item pages. The template puts spacer rows inside the items table thead and tfoot, so every continuation page gets ~80px above and ~60px below the rows instead of items butting up against the page edge. Chromium ignores @page :first and :nth() on continuation pages, so the spacers have to live inside the table for them to auto-repeat.

Tested at 30, 60 and 120 distinct line items.

## Removed

- src/service/pdf/invoice-pdf-service.ts (the LaTeX service)
- test/unit/service/invoice-pdf-service.ts (replaced with an HTML test)
- emptyIdentity and subTransactionRowToProduct from src/helpers/pdf.ts (no other callers; the PDF_VAT_* constants and UNUSED_PARAM stay for the six remaining LaTeX PDFs)

## Not removed yet

pdf-generator-client is still in package.json since six other PDFs use it. Removing it is the final PR in the series, per PLAN-invoice-pdf-migration.md section 3.

#926

## Types of changes

- New feature
- Style

---

## PR Checklist

- [x] **Test Coverage**: New functionality has test coverage and all tests pass (pnpm test: 2275 passed, 27 skipped, 0 failed)
- [x] **Documentation**: New code has TypeDoc comments
- [x] **Database Changes**: N/A

## Additional Notes

Smoke tested end-to-end against a running backend with the abc.docker-registry.gewis.nl/eou/pdf-compiler:latest container running locally. The rendered PDF matches the F6c reference output.


src/files/templates/bac-letterhead.ts
RubenLWF, line 1: question:

Is this indeed the way we want to store this data?

This means that if we want to change, for example, the payment term, we need to make a PR, rather than changing something in the DB or ENV variables.

RubenLWF (changes requested):

src/files/templates/bac-letterhead.ts
JustSamuel: > Is this indeed the way we want to store this data?
This PR keeps parity, it was not part of the scope to change the variables.

We can do this later (part of invoices? env? db?).

RubenLWF (approved):

PR #932, opened 2026-05-22
Upgrade to TypeORM 1.0

JustSamuel:
## Summary

- Bump `typeorm` from 0.3.29 to 1.0.0. Raise the TS and vitest swc target from ES2021 to ES2023, which is the new minimum.
- Pin `useDefineForClassFields: false` in tsconfig.json. ES2023 flips its default to `true`, so tsc starts emitting `this.relation = undefined` for declared-but-uninitialized relation fields. That breaks TypeORM entity hydration: `Permission.role` becomes a literal `undefined` property and shows up in property iteration. CI's JS-compiled test path caught it via `default-roles`. The TS-source vitest pass didn't, because swc already had this option set.
- Apply `@typeorm/codemod v1` across `src/`, `test/`, `cli/` to convert string-array `relations`/`select` find options to object syntax and rename `findByIds`/`findOneById`/`exist`.
- Add `invalidWhereValuesBehavior` settings in the DataSource so existing filter helpers keep working under the new defaults.

Closes #929.

## Behavioral changes I had to handle

Two 1.0 default changes broke things in `src/database/database.ts`. I pinned both back to 0.3 behavior instead of rewriting hundreds of call sites.

`undefined` in `where` used to be silently skipped. In 1.0 it throws. Plenty of services do `createdAt: createFilterWhereDate(from, till)` directly inside `where` and rely on the key dropping out when both dates are missing. Without the setting that's 175 test failures. With `invalidWhereValuesBehavior.undefined = 'ignore'` it's back to working.

`null` is similar, except some of the offenders are TypeORM itself. The internal soft-delete handling emits `deletedAt: null` when joining to entities with `@DeleteDateColumn` (e.g. `EventShift`), so the new default throws on queries that worked fine in 0.3. Setting `null = 'sql-null'` translates it to SQL `IS NULL`, which is what it's supposed to mean in that context anyway.

The other documented behavioral change, non-nullable relations now using INNER JOIN, I left alone. FK constraints rule out orphans on these relations and nothing in the test suite tripped.

## What I audited and didn't adopt

`await using` on QueryRunner. The 1.0 release notes claim support, but `QueryRunner.d.ts` doesn't actually declare `[Symbol.asyncDispose]`. tsc rejects `await using queryRunner = ds.createQueryRunner()` with TS2851. Worth revisiting once the types catch up; for now `try { ... } finally { release() }` stays.

`returning` on `update()`/`upsert()` needs a `RETURNING` clause. Prod is MySQL, no dice.

`ifExists` on the various drop methods only matters for new migrations. Editing already-applied ones is pointless churn.

The PostgreSQL / SAP HANA / MongoDB / Spanner / Aurora improvements are for engines we don't run.

The remaining bits (SQLite `jsonb`, `@Exclusion deferrable`, `increment`/`decrement` typing, DataSource-level `isolationLevel`, `valuesFromSelect()`) don't have a call site that would benefit.

## Test plan

- [x] `pnpm exec tsc --noEmit` clean
- [x] `pnpm lint` clean
- [x] `pnpm test` -- 2368 pass, 27 skipped on the rebased branch. (175 failed before the compat setting; the `default-roles` JS-compiled regression was fixed by the `useDefineForClassFields` pin.)
- [x] `pnpm run schema` -- in-memory SQLite bootstrap works
- [x] `pnpm run migrate` against MySQL -- ran via CI's `coverage-mariadb` job (passed, 5m9s)

Upgrade guide: https://typeorm.io/docs/releases/1.0/upgrading-from-0.3

### DIFF
--- a/src/controller/event-controller.ts
+++ b/src/controller/event-controller.ts
@@ -335,7 +335,10 @@ export default class EventController extends BaseController {
 
     try {
       const parsedId = Number.parseInt(id, 10);
-      const event = await Event.findOne({ where: { id: parsedId }, relations: ['answers', 'shifts'] });
+      const event = await Event.findOne({ where: { id: parsedId }, relations: {
+        answers: true,
+        shifts: true,
+      } });
       if (event == null) {
         res.status(404).send();
         return;
@@ -374,7 +377,9 @@ export default class EventController extends BaseController {
     let shiftId = Number.parseInt(rawShiftId, 10);
     let userId = Number.parseInt(rawUserId, 10);
     try {
-      const answer = await EventShiftAnswer.findOne({ where: { eventId, shiftId, userId }, relations: ['event'] });
+      const answer = await EventShiftAnswer.findOne({ where: { eventId, shiftId, userId }, relations: {
+        event: true,
+      } });
       if (answer == null) {
         res.status(404).send();
         return;
@@ -427,7 +432,9 @@ export default class EventController extends BaseController {
     let shiftId = Number.parseInt(rawShiftId, 10);
     let eventId = Number.parseInt(rawEventId, 10);
     try {
-      const answer = await EventShiftAnswer.findOne({ where: { eventId, shiftId, userId }, relations: ['event'] });
+      const answer = await EventShiftAnswer.findOne({ where: { eventId, shiftId, userId }, relations: {
+        event: true,
+      } });
       if (answer == null) {
         res.status(404).send();
         return;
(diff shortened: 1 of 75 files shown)


src/controller/event-controller.ts
```diff
@@ -216,7 +216,7 @@ export default class EventController extends BaseController {
     let params: CreateEventParams;
     try {
       params = {
-        ...await parseUpdateEventRequestParameters(req),
+        ...(await parseUpdateEventRequestParameters(req)),
```
RubenLWF: question:

why is this change required?

RubenLWF (approved):
Looking good.

Please fix the merge conflicts :)

src/controller/event-controller.ts
JustSamuel: Reverted both occurrences back to the original form.

PR #952, opened 2026-06-09
Migrate fine report PDF to pdf-compiler

JustSamuel:
# Description

Migrates the fine report PDF off the old LaTeX `pdf-generator-client` path onto `pdf-compiler`. Second of the PDF migrations tracked in #926; the invoice was the first (#931).

`FineReportPdfService` was LaTeX, generating through `client.generateFineReport`. It is now an HTML service that renders a new `fine-report.html.ts` template through `pdf-compiler`. A fine report is an unstored aggregate report, so it follows the inactive administrative cost report rather than the invoice: it extends `HtmlUnstoredPdfService` and reuses the shared `base.html` letterhead, which gives it the same green BAC styling as the invoice. The VAT math from the old service is unchanged (handed out minus waived, with the 21% high VAT split out). The `GET /fines/report/pdf` route, its RBAC, and the controller are untouched -- only the service behind the entity's `pdfService` field changed, so the API surface stays the same.

`pdf-generator-client` and the `PDF_VAT_*` constants stay, since the other LaTeX PDFs in #926 still use them. No database changes.

Verified by rendering the report against a live `pdf-compiler` container: `GET /fines/report/pdf` returned a valid PDF with the BAC letterhead, the Report Period and Total cards, the Handed out line, and the 21% VAT breakdown. `tsc`, eslint, the new html/service unit tests, and the debtor-controller suite all pass.

Part of #926.

## Types of changes

---

## ✅ PR Checklist

## 🔗 Additional Notes
The conditional "Waived" row only renders when fines were waived in the period (covered by the html/service unit tests). The seeded dev data has a handed-out fine but nothing waived, so the live-rendered sample shows the handed-out-only layout.


CodeNamedRobin (approved):
Looks good to me!

PR #954, opened 2026-06-11
Migrate seller payout report PDF to pdf-compiler

JustSamuel:
# Description

Migrates the seller payout / disbursement report PDF off the LaTeX pdf-generator-client path onto pdf-compiler. Third of the PDF migrations tracked in #926, after the invoice (#931) and the fine report (#952).

A disbursement is the same document as the invoice, just with different header and info-box text, so this also extracts the shared layout. The invoice's cover + line-items skeleton (band, recipient block, meta-list, total headline, VAT specification, footer, and the line-items page) moves into a new document.html.ts. invoice.html.ts and seller-payout.html.ts are now thin wrappers that pass their own band label, meta rows, note text and line items. The invoice output is unchanged and its tests still pass.

The seller payout renders as a cover page with the "Seller Disbursement" header, the account, the payout meta (Payout number, Account, Start date, End date, Customer number), a Description section, a "Payout including VAT" headline and the disbursement note ("This is a payout of your balance ... responsible for payment of VAT in the amount of ..."), then a line-items page listing the products sold. Amounts use nl-NL formatting (EUR 8,70).

SellerPayoutPdfService now extends HtmlPdfService and builds numeric line items and VAT bands straight from the sales report. It is still a stored PDF, so the SellerPayout entity keeps its pdf relation and pdfConstructor and the getOrCreatePdf caching is unchanged. The GET /seller-payouts/{id}/report/pdf route, its RBAC and the controller handler are untouched.

Verified end to end against a live pdf-compiler container: GET /seller-payouts/1/report/pdf on the dev seed stored a valid two-page PDF. tsc, eslint, the new html/service unit tests, the invoice unit tests and the seller-payout-controller suite all pass.

Part of #926.

## Types of changes

---

## ✅ PR Checklist

## 🔗 Additional Notes
No new dependencies and no DB migration. The shared document.html.ts is now used by both the invoice and the seller payout; future PDFs in #926 with a recipient + line items (e.g. payout request) can reuse it too. pdf-generator-client and the PDF_VAT_* constants stay for the PDFs not yet migrated.

### DIFF
--- a/src/html/seller-payout.html.ts
+++ b/src/html/seller-payout.html.ts
@@ -1,0 +1,101 @@
+/**
+ *  SudoSOS back-end API service.
+ *  Copyright (C) 2026 Study association GEWIS
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as published
+ *  by the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ *  @license
+ */
+
+/**
+ * Seller Payout / disbursement PDF. The same "F6c" document as the Invoice,
+ * with a "Seller Disbursement" band label, the payout meta rows and the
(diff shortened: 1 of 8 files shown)


src/html/seller-payout.html.ts
```diff
@@ -0,0 +1,109 @@
...
+import {
+  createDocumentPdf,
+  escapeHtml,
+  fmt,
+  IDocumentLineItem,
+  IDocumentVatBand,
+} from './document.html';
+import { BAC } from '../files/templates/bac-letterhead';
+
+const MUTED = '#6B6B6B';
```
RubenLWF: Question:

why do we define this in both the controller and the HTML?

JustSamuel: Addressed -- it was a leftover from the extraction: both wrappers duplicated the same inline-styled section markup, so each redefined the color constant. Pulled that markup into a shared subjectSection(label, value) helper in document.html.ts, so MUTED now only lives there.

